PR #10, opened 2020-12-22
adding export to csv

sinisaos:
@dantownsend this is my try on exporting to csv. Probably it be slow for large and complex tables due to passing total number of rows to PiccoloCRUD __page_size param , but it's just idea.

### DIFF
--- a/admin_ui/src/components/CSVButton.vue
+++ b/admin_ui/src/components/CSVButton.vue
@@ -21,11 +21,6 @@ import * as i from "../interfaces"
 
 export default {
     props: ["tableName"],
-    data() {
-        return {
-            tableName: this.tableName,
-        }
-    },
     components: {
         downloadexcel,
     },
@@ -33,16 +28,18 @@ export default {
         // Export data as csv from json:
         async fetchExportedRows() {
             const params = this.$store.state.filterParams
-            const tableName = this.$store.state.currentTableName
             const sortBy = this.$store.state.sortBy
             if (sortBy) {
                 let prefix = sortBy.ascending ? "" : "-"
                 params["__order"] = prefix + sortBy.property
             }
             // Get the row counts:
-            const response = await axios.get(`api/tables/${tableName}/count/`, {
-                params,
-            })
+            const response = await axios.get(
+                `api/tables/${this.tableName}/count/`,
+                {
+                    params,
+                }
+            )
             const data = response.data as i.RowCountAPIResponse
             const localParams = { ...params }
 
@@ -56,7 +53,7 @@ export default {
                 for (let i = 1; i < pages + 1; i++) {
                     localParams["__page"] = i
                     const response = await axios.get(
-                        `api/tables/${tableName}/?__readable=true`,
+                        `api/tables/${this.tableName}/?__readable=true`,
                         {
                             params: localParams,
                         }
@@ -68,20 +65,10 @@ export default {
                 console.log(error.response)
             }
         },
-        async startDownload() {
-            const params = this.$store.state.filterParams
-            const tableName = this.$store.state.currentTableName
-            const response = await axios.get(`api/tables/${tableName}/count/`, {
-                params,
-            })
-            const data = response.data as i.RowCountAPIResponse
-            if (data.count > 10000) {
-                alert(
-                    "Number of rows is really high. Do you want to start download your data."
-                )
-            } else {
-                this.startDownload
-            }
+        startDownload() {
+            alert(
+                "Your data will begin downloading. Large data sets may take a few seconds."
+            )
         },
         finishDownload() {
             alert("Your data is ready.")
--- a/admin_ui/src/components/CSVButton.vue
+++ b/admin_ui/src/components/CSVButton.vue
@@ -1,0 +1,94 @@
+<template>
+    <a class="button">
+        <downloadexcel
+            :before-finish="finishDownload"
+            :before-generate="startDownload"
+            :fetch="fetchExportedRows"
+            :name="tableName + '.csv'"
+            type="csv"
+        >
+            <span>
+                <font-awesome-icon icon="file-csv" />Export as CSV
+            </span>
+        </downloadexcel>
+    </a>
+</template>
+
+<script lang="ts">
+import axios from "axios"
+import downloadexcel from "vue-json-excel"
+import * as i from "../interfaces"
+
+export default {
+    props: ["tableName"],
+    data() {
+        return {
+            tableName: this.tableName,
+        }
+    },
+    components: {
+        downloadexcel,
+    },
+    methods: {
+        // Export data as csv from json:
+        async fetchExportedRows() {
+            const params = this.$store.state.filterParams
+            const tableName = this.$store.state.currentTableName
+            const sortBy = this.$store.state.sortBy
+            if (sortBy) {
+                let prefix = sortBy.ascending ? "" : "-"
+                params["__order"] = prefix + sortBy.property
+            }
+            // Get the row counts:
+            const response = await axios.get(`api/tables/${tableName}/count/`, {
+                params,
+            })
+            const data = response.data as i.RowCountAPIResponse
+            const localParams = { ...params }
+
+            localParams["__page"] = data.count
+            // Set higher __page_size param to have fewer requests to api:
+            localParams["__page_size"] = 1000
+            const pages = Math.ceil(data.count / localParams["__page_size"])
+            const exportedRows = []
+
+            try {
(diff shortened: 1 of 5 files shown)


sinisaos:
@dantownsend changes in the second commit has much better performace. I hope that is better solution.

dantownsend:
@sinisaos Great - thanks for tackling this. I'll give it a proper look over tonight. CSV exports is something people find really useful, so this will be a valuable feature.

sinisaos:
@dantownsend Great, thanks for your time. Small number of rows is not a problem. I able to get exported 50000 rows of table with 5 column with foreign key in 10 seconds on localhost with `uvicorn app:app --port 8000 --host 0.0.0.0` without workers (I dont know how many rows is good result for exporting?). With previous commit browser crashes on 50000 rows (because I loaded to much data) but now I use loop to get data from api and things look much better.

dantownsend:
@sinisaos It's a good start - thanks again for tackling this. Here are some suggestions:

## Separate component?

It could be moved it into a new component called something like `CSVButton` - I can imagine `RowListing` getting quite big over time.

## JS formatting

We need to synchronise our prettier settings - as part of the diff is adding semicolons.

I use VSCode - in settings.json I have the following:

```
    "vetur.format.options.tabSize": 4,
    "vetur.format.options.useTabs": false,
    "vetur.format.enable": true,
    "vetur.format.defaultFormatter.js": "prettier",
    "vetur.format.defaultFormatterOptions": {
        "prettier": {
            "semi": false
        },
        "prettyhtml": {
            "printWidth": 100,
            "singleQuote": false,
            "wrapAttributes": true,
            "sortAttributes": true
        }
    },
```

Which editor do you use? It could also be because we’re using different versions of Prettier / Vetur.

## Params object

The `params` object is being manipulated directly - it’s best to avoid manipulating objects from the store in this way. Instead you can make a local copy of the params, like this:

```
const localParams = {…params}
```

This means you won’t have to reset the params in `finishDownload` either.

## startDownload

I think it would be good if the row count was fetched here, and an alert shown if the number of rows is really high - if so they need to confirm before proceeding.

admin_ui/src/views/RowListing.vue
```diff
@@ -183,95 +225,150 @@ export default Vue.extend({
...
+            const response = await axios.get(`api/tables/${tableName}/count/`, {
+                params,
+            });
+            const data = response.data as i.RowCountAPIResponse;
+
+            params["__page"] = data.count;
+            // set greater __page_size param to have fewer requests to api
+            params["__page_size"] = 1000;
+            const pages = Math.ceil(data.count / params["__page_size"]);
+            const expotredRows = [];
```
dantownsend, line 286: Small typo here => `exportedRows`

sinisaos:
@dantownsend hope I didn't miss something from proposed suggestions :)

admin_ui/src/components/CSVButton.vue
```diff
@@ -0,0 +1,94 @@
...
+import downloadexcel from "vue-json-excel"
+import * as i from "../interfaces"
+
+export default {
+    props: ["tableName"],
+    data() {
+        return {
+            tableName: this.tableName,
+        }
+    },
```
dantownsend, line 28: Just defining `tableName` as a prop should be enough - `data` can be removed.

admin_ui/src/components/CSVButton.vue
```diff
@@ -0,0 +1,94 @@
...
+    components: {
+        downloadexcel,
+    },
+    methods: {
+        // Export data as csv from json:
+        async fetchExportedRows() {
+            const params = this.$store.state.filterParams
+            const tableName = this.$store.state.currentTableName
```
dantownsend, line 36: You should be able to get `tableName` from `this.tableName` instead of reading it from the store.

admin_ui/src/components/CSVButton.vue
```diff
@@ -0,0 +1,94 @@
...
+                console.log(error.response)
+            }
+        },
+        async startDownload() {
+            const params = this.$store.state.filterParams
+            const tableName = this.$store.state.currentTableName
+            const response = await axios.get(`api/tables/${tableName}/count/`, {
+                params,
+            })
+            const data = response.data as i.RowCountAPIResponse
```
dantownsend, line 77: I just realised, the row count is available in the store as `this.$store.state.rowCount`. It's calculated when the page loads, so if the page has been open a while, it could be slightly out of date, but probably still OK to use.

If we decide to keep with the current approach, and fetch the latest count, it might be worth storing the count as a data attribute - i.e. `this.rowCount = response.count`, so it can be reused between `startDownload` and `fetchExportedRows`. It just saves one API call.

admin_ui/src/components/CSVButton.vue
```diff
@@ -0,0 +1,94 @@
...
+            if (data.count > 10000) {
+                alert(
+                    "Number of rows is really high. Do you want to start download your data."
+                )
```
dantownsend, line 81: We need to change this to:

```javascript
        if confirm(
                "The number of rows is really high. Do you want to start downloading your data?"
         ) {
             this.startDownload
         }
```
So when they click OK, it starts downloading.

dantownsend:
@sinisaos Thanks for making those changes - it's getting very close! Just a couple more suggestions from me :)

admin_ui/src/components/CSVButton.vue
sinisaos: But data is start download when click export button and `confirm()` doesn't change anything. I think that we should go simpler

```
startDownload() {
    alert("Your data is start downloading. For large dataset download may take few seconds.")
},
finishDownload() {
	alert("Your data is ready.")
},
```

With that we nicely inform user that large dataset need few seconds to download and we save one API call because don't need rows count in `startDownload()`.  What do you think about that?

sinisaos:
@dantownsend an this is whole example of `CSVButton.vue`

```
<template>
    <a class="button">
        <downloadexcel
            :before-finish="finishDownload"
            :before-generate="startDownload"
            :fetch="fetchExportedRows"
            :name="tableName + '.csv'"
            type="csv"
        >
            <span>
                <font-awesome-icon icon="file-csv" />Export as CSV
            </span>
        </downloadexcel>
    </a>
</template>

<script lang="ts">
import axios from "axios"
import downloadexcel from "vue-json-excel"

export default {
    props: ["tableName"],
    components: {
        downloadexcel,
    },
    methods: {
        // Export data as csv from json:
        async fetchExportedRows() {
            const params = this.$store.state.filterParams
            const sortBy = this.$store.state.sortBy
            if (sortBy) {
                let prefix = sortBy.ascending ? "" : "-"
                params["__order"] = prefix + sortBy.property
            }
            // Get the row counts:
            const response = await axios.get(`api/tables/${this.tableName}/count/`, {
                params,
            })
            const data = response.data as i.RowCountAPIResponse
            const localParams = { ...params }

            localParams["__page"] = data.count
            // Set higher __page_size param to have fewer requests to api:
            localParams["__page_size"] = 1000
            const pages = Math.ceil(data.count / localParams["__page_size"])
            const exportedRows = []

            try {
                for (let i = 1; i < pages + 1; i++) {
                    localParams["__page"] = i
                    const response = await axios.get(
                        `api/tables/${this.tableName}/?__readable=true`,
                        {
                            params: localParams,
                        }
                    )
                    exportedRows.push(...response.data.rows)
                }
                return exportedRows
            } catch (error) {
                console.log(error.response)
            }
        },
        async startDownload() {
            alert("Your data is start downloading. Downloading a large set of data may take a few seconds.")
        },
        finishDownload() {
            alert("Your data is ready.")
        },
    },
}
</script>
```

admin_ui/src/components/CSVButton.vue
dantownsend: Ah, I see - didn't realise that.

dantownsend:
> @dantownsend an this is whole example of `CSVButton.vue`

@sinisaos Yeah, that looks good. Just some very minor things I'd suggest, then we'll get it merged. How about changing the alert text to:

`alert("Your data will begin downloading. Large data sets may take a few seconds.")`

It's just slightly more concise.

And the `startDownload` method no longer needs to be async.

sinisaos:
@dantownsend First sorry for putting whole code in comment, but I wasn't at home and I couldn't have done otherwise . `startDownload()` is a typo and I fix it. Also I replace alert with yours and then commit changes. I'm glad that you like this simpler way.

dantownsend:
Cool - nice work 👍

sinisaos:
Thanks. I'm glad I can help.

PR #11, opened 2021-01-02
making bulk delete option

sinisaos:
Hi Daniel
This is my attempt to make bulk delete rows on `currentPage` like in Django Admin. Suggestions and help is welcome. Bulk delete button is separate component but checkboxes is in `RowListing.vue` with logic on page change, table change etc.

### DIFF
--- a/admin_ui/src/App.vue
+++ b/admin_ui/src/App.vue
@@ -21,7 +21,7 @@ import MessagePopup from "./components/MessagePopup.vue"
 export default Vue.extend({
     components: {
         AboutModal,
-        MessagePopup
+        MessagePopup,
     },
     computed: {
         darkMode() {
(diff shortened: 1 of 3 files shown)


dantownsend:
@sinisaos This is a great start, and probably the number 1 most important missing feature at the moment.

It does raise some issues though - there are quite a few buttons now, and it's starting to break on mobile. Some of the button names might have to be shortened.

Screenshot 2021-01-02 at 22 22 12

Also, it's having to call the delete endpoint once per row. There is a bulk deletion endpoint in ``PiccoloCRUD``, but it isn't enabled on Piccolo Admin, and doesn't currently allow a bulk delete which filters on certain row IDs, which would be required for this.

I was also thinking that the Piccolo admin API could be converted to use ``FastAPIWrapper`` instead, as having the API docs would be useful, but this could be done separately in the future.

What I'm thinking, is merging this into a feature branch, and giving you write permissions to it. I can then play around with it a bit - adding support for bulk deleting certain row IDs, for example.

What do you think?

sinisaos:
@dantownsend Of course. This example is just an idea but also the best of me. You will do it much better :) .

PR #15, opened 2021-01-05
Interval widget

sinisaos:
@dantownsend Here is all changes made. You can easily see changes, test it and review it. Cheers.


dantownsend:
@sinisaos Brilliant, thanks. I'll get this merged in ASAP.

I've been playing around with some different UI ideas for making the buttons work better on smaller screens. I was thinking of adding a dropdown for bulk actions, but for now will just tweak the button styles on mobile.

Screenshot 2021-01-05 at 17 25 13

I thought about adding an API endpoint for bulk deletion, but it needs careful consideration, so will leave it for now.

PR #17, opened 2021-01-10
component for dynamically change the page size

sinisaos:
Hi Daniel
Here is component for dynamically change page size in admin UI. Cheers.

### DIFF
--- a/admin_ui/src/components/ChangePageSize.vue
+++ b/admin_ui/src/components/ChangePageSize.vue
@@ -1,0 +1,41 @@
+<template>
+    <p>
+        Show
+        <select
+            id="page_size"
+            v-model="selectedPageSize"
+            v-on:click.prevent="$emit('triggered')"
+        >
+            <option
+                v-bind:key="option"
+                v-for="option in pageOptions"
+            >{{ option }}</option>
+            {{ pageSize }}
+        </select>
+        results
+    </p>
+</template>
+
+<script>
+export default {
+    data() {
+        return {
+            selectedPageSize: 5,
+            pageOptions: [5, 10, 25, 50, 100],
+        }
+    },
+    computed: {
+        pageSize() {
+            this.$store.state.pageSize = this.selectedPageSize
+            return this.$store.state.pageSize
+        },
+    },
+}
+</script>
+
+<style lang="less" scoped>
+select#page_size {
+    width: 3.3rem;
+    margin-bottom: -5px;
+}
+</style>
--- a/admin_ui/src/store.ts
+++ b/admin_ui/src/store.ts
@@ -98,10 +98,12 @@ export default new Vuex.Store({
                     params
                 }
             )
+            params['__page_size'] = context.state.pageSize
+
             const data = response.data as i.RowCountAPIResponse
             context.commit('updateRowCount', data.count)
-            context.commit('updatePageSize', data.page_size)
-            if (data.count < data.page_size) {
+            context.commit('updatePageSize', params['__page_size'])
+            if (data.count < params['__page_size']) {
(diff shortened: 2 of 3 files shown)


admin_ui/src/components/ChangePageSize.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+    <p>
+        Show
+        <select
+            id="page_size"
+            v-model="selectedPageSize"
+            v-on:click.prevent="$emit('triggered')"
```
dantownsend, line 7: This should be `v-on:changed` instead.

admin_ui/src/components/ChangePageSize.vue
```diff
@@ -0,0 +1,41 @@
...
+export default {
+    data() {
+        return {
+            selectedPageSize: 5,
+            pageOptions: [5, 10, 25, 50, 100],
+        }
+    },
+    computed: {
+        pageSize() {
+            this.$store.state.pageSize = this.selectedPageSize
```
dantownsend, line 29: It's best to avoid setting state in computed properties.

dantownsend:
@sinisaos This is a great start - thanks for tackling this. I'm thinking of moving the select so it's at the bottom, next to the pagination. I've added a couple of code suggestions, but I'm happy to merge it into a feature branch, and pick these up - it's up to you. Thanks again.

dantownsend:
Here's a very rough idea of what the select could look like at the bottom of the page.

Screenshot 2021-01-10 at 23 51 26

sinisaos:
@dantownsend Of course :). You can do whatever you need. I have two questions regarding the roadmap. 

1. Creating users and changing passwords via the UI - we already have create and edit Piccolo User. On creating password are hashed, and for editing user we can add hashing to PiccoloCrud. Something like this in PiccoloCrud `endpoints.py`.

```
# import at beginning of file
from piccolo.apps.user.tables import BaseUser

# rest of code

async def _patch_single(self, row_id: int, data: t.Dict[str, t.Any]) -> Response:
        """
        Patch a single row.
        """
        cleaned_data = self._clean_data(data)

        if "password" in cleaned_data:
            cleaned_data["password"] = BaseUser.hash_password(cleaned_data["password"])

# rest of code

```
Everything is already working in the admin UI. Maybe we need to make the errors more user friendly (instead of `Internal server error ` something like `User already exists` or `User with that email or username already exists`). Sorry if I miss your point

2. Be able to specify a site name in create_admin, so the UI shows MySite Admin in the nav bar, rather than the default Piccolo Admin - why you want to change that. I think it is Piccolo Admin a good name and more or less all administrator interfaces have a framework name inside (Django Administration, Flask-Admin, FastAPI-Admin etc).

dantownsend:
@sinisaos The password thing is quite tricky. It's hashed when stored in the database, and when you see it in the admin, you see this hashed value. The intention for adding a `Secret` column type was that when serialising them, the value would be replaced with some placeholder, so they're not sent over the wire at all.

Ideally, the Piccolo Admin will then identify that it's a `Secret` field, and will display it as a password field. If doing a PATCH, `PiccoloCRUD` compares whether the new value matches the placeholder, if not it should hash and save it, as you outlined.

For point 2, my idea was you could change it to be 'My Company Admin' or something. I used to work for a digital agency, and some customers would like to see their logo or name in the admin for some reason. It's not a big deal.

What do you think are the main improvements which can be made to the admin?

sinisaos:
@dantownsend Thanks for explaining things to me and sorry for my misunderstanding the problem.

dantownsend:
@sinisaos No worries!

PR #20, opened 2021-01-15
ability to change site name instead of default Piccolo

sinisaos:
Hi Daniel
here is my attempt to add ability to change site name instead of the default `Piccolo Admin` in `create_admin`. Cheers.

### DIFF
--- a/admin_ui/src/components/NavBar.vue
+++ b/admin_ui/src/components/NavBar.vue
@@ -12,7 +12,8 @@
 
             <h1>
                 <router-link to="/">
-                    <font-awesome-icon icon="tools" />Piccolo Admin
+                    <font-awesome-icon icon="tools" />
+                    {{ siteName }} Admin
                 </router-link>
             </h1>
 
@@ -53,7 +54,7 @@ export default Vue.extend({
     data() {
         return {
             showSidebar: false,
-            showDropdown: false
+            showDropdown: false,
         }
     },
     computed: {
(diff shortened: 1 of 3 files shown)


dantownsend:
@sinisaos Great job! Thanks for this.

I have a couple of suggestions, but happy to pick these up if you want.

piccolo_admin/endpoints.py
```diff
@@ -58,8 +58,10 @@ def __init__(
         read_only: bool = False,
         rate_limit_provider: t.Optional[RateLimitProvider] = None,
         production: bool = False,
+        sitename: str = "Piccolo",
```
dantownsend, line 61: How about `site_name` instead?

piccolo_admin/endpoints.py
```diff
@@ -149,6 +151,14 @@ def __init__(
...
                             ),
                         ),
+                        Mount(
+                            path="/sitename/",
+                            app=auth_middleware(
+                                Router(
+                                    [Route(path="/", endpoint=self.get_sitename)]
+                                )
+                            ),
+                        ),
```
dantownsend, line 161: Rather than having a separate endpoint, we could return the `site_name` as part of the meta endpoint?

admin_ui/src/components/NavBar.vue
```diff
@@ -12,7 +12,8 @@
 
             <h1>
                 <router-link to="/">
-                    <font-awesome-icon icon="tools" />Piccolo Admin
+                    <font-awesome-icon icon="tools" />
+                    {{ siteName }} Admin
```
dantownsend, line 16: How do you feel about removing the word 'Admin'? The default value coming back from the API would then have to be 'Piccolo Admin' instead of 'Piccolo'.

I can imagine someone just wanting 'My Business' rather than 'My Business Admin', if the name is long.

Likewise, if the website is for a non-English audience, they might want to translate the word 'Admin'. Most of the admin is hardcoded English anyway at the moment, but hopefully one day it'll be more multi-lingual.

sinisaos:
@dantownsend Of course :) . Can you close the resolved issues and update roadmap with finished items. Cheers.

dantownsend:
@sinisaos Cool, will do

PR #25, opened 2021-01-24
minor meta fix

sinisaos:
@dantownsend Minor fixes for a custom site name because metadata doesn't display properly until the page refreshes. You can see that in [demo](https://demo1.piccolo-orm.com/#/login?nextURL=%2F), version in the modal is `Unknown` until you refresh the page. With these changes, everything should be fine with the first page view (at least that's the case with me), but you check it out. Cheers.

### DIFF
--- a/admin_ui/src/modules/meta.ts
+++ b/admin_ui/src/modules/meta.ts
@@ -8,18 +8,18 @@ export default {
         piccoloAdminVersion: 'Unknown'
     },
     mutations: {
-        updateSiteName(state, value: Object) {
+        updateSiteName(state, value: string) {
             state.siteName = value
         },
-        updatePiccoloAdminVersion(state, value: Object) {
+        updatePiccoloAdminVersion(state, value: string) {
             state.piccoloAdminVersion = value
         },
     },
     actions: {
         async fetchMeta(context) {
             const response = await axios.get(`${BASE_URL}meta/`)
-            this.commit('updateSiteName', response.data.site_name)
-            this.commit('updatePiccoloAdminVersion', response.data.piccolo_admin_version)
+            context.commit('updateSiteName', response.data.site_name)
+            context.commit('updatePiccoloAdminVersion', response.data.piccolo_admin_version)
         }
     }
 }
--- a/admin_ui/src/views/Home.vue
+++ b/admin_ui/src/views/Home.vue
@@ -22,6 +22,9 @@ export default {
     mounted() {
         this.$store.commit("updateCurrentTablename", "")
     },
+    async created() {
+        await this.$store.dispatch("fetchMeta")
+    },
 }
 </script>
 


admin_ui/src/modules/meta.ts
```diff
@@ -8,18 +8,18 @@ export default {
...
         },
     },
     actions: {
         async fetchMeta(context) {
             const response = await axios.get(`${BASE_URL}meta/`)
-            this.commit('updateSiteName', response.data.site_name)
-            this.commit('updatePiccoloAdminVersion', response.data.piccolo_admin_version)
+            context.commit('updateSiteName', response.data.site_name)
+            context.commit('updatePiccoloAdminVersion', response.data.piccolo_admin_version)
         }
```
dantownsend, line 23: Good catch - not sure how I missed these issues.

admin_ui/src/views/Home.vue
```diff
@@ -22,6 +22,9 @@ export default {
     mounted() {
         this.$store.commit("updateCurrentTablename", "")
     },
+    async created() {
+        await this.$store.dispatch("fetchMeta")
+    },
```
dantownsend, line 27: It's also dispatching this in the create method of App.vue, so the API is being called twice. I wonder if the problem was just related to the fixes you made in `meta.ts`, and we don't need this additional API call? Does it still work for you locally without it?

dantownsend:
@sinisaos Thanks for this

admin_ui/src/views/Home.vue
sinisaos: It's weird but works only if api call is from `Home.vue`. I completely remove api call from `App.vue` and  
leave api call from `Home.vue`, and everything work correctly.

dantownsend: The only problem with just calling it in `Home.vue`, is if someone opens the website on a different page, it wouldn't call it. It will also get called every time `Home.vue` gets opened.

In theory, `App.vue` is the best place for it - as it should always get called, and only gets called once, but there could be a bug which is stopping it from working. Which browser do you use?

sinisaos: Chromium Version 87.0.4280.66(64-bit) and Mozilla Firefox 84.0.2 (64-bit). My OS is Linux Mint 19.1. Is this working for you just with changes in `meta.ts`?

dantownsend: I was able to replicate it when opening in an incognito window - but only some of the time. It's very strange. Probably means there's a race condition somewhere. I'll play around a bit more.

dantownsend: I think I know what the issue is now. If you load the admin and you're not logged in, the app tries to fetch the meta, but if the user isn't logged in it fails. This is why adding it to the `Home.vue` fixed it.

dantownsend: I'll add the API call to `Login.vue` instead - so when the user logs in, it fetches the meta. That should fix it!

sinisaos: Well done! I tried to add an API call to `Login.vue` and everything works great.

PR #31, opened 2021-02-03
FastAPI backend

dantownsend:
Using `FastAPIWrapper` instead of `PiccoloCRUD` to create API docs.

Slight API refactor - login and logout endpoints are now served at `/auth` instead of `/api`.


dantownsend:
@sinisaos I think I've finally finished the FastAPI backend for Piccolo admin. I've built on top of your work. How does this look to you?

sinisaos:
@dantownsend Looks great :). Cheers.

PR #32, opened 2021-02-05
small path fix for auth

sinisaos:
Small path fix for auth. `/auth/login `and `/auth/logout` produce `404` error. It should be `./auth/login` and `./auth/logout/`.


dantownsend:
@sinisaos Good catch - bad mistake on my part.

I was only testing the admin when hosted at the root.

sinisaos:
@dantownsend Neither did I notice yesterday until I tried it today. Everything else works great.

PR #33, opened 2021-02-13
small underscore fix

sinisaos:
Small fix when column name contains underscore in Schema definition. 

```
class Task(Table):
    """
    An Task table.
    """

    name = Varchar()
    description = Text()
    completed = Boolean(default=False)
    task_user = ForeignKey(references=BaseUser) # <- like this underscore 
```

With this change column content will display correctly in piccolo admin after creation.


dantownsend:
@sinisaos Thanks for this.

sinisaos:
@dantownsend No problem. Cheers.

Issue #30, opened 2021-02-03
Improve foreign key selector when dealing with lots of rows


dantownsend:
The foreign key selectors are `select` HTML elements. These are used in the sidebar for filtering, and for changing foreign key values in detail views.

It works fine when the number of values is relatively small. If there are thousands of values, it could become less useful.

<img width="320" alt="Screenshot 2021-02-03 at 20 12 22" src="https://user-images.githubusercontent.com/350976/106803444-22b14780-665c-11eb-8389-bf7406f4f0ba.png">

A search box might be better in these situations.

`PiccoloCRUD` has an `ids` endpoint - `/api/tables/<tablename>/ids` which returns a mapping of row ID to a readable value. For example:

```json
{
    "1": "Peter Jackson",
    "2": "George Lucas",
    "3": "J.J. Abrams",
    "4": "Gareth Edwards",
    "5": "Irvin Kershner",
    "6": "Richard Marquand",
    "7": "Rian Johnson",
    "8": "Ron Howard"
}
```
Try it out [here](https://demo1.piccolo-orm.com/api/tables/director/ids/).

The endpoint could be modified to return data in this format:

```json
{
    "total_count":  8,
    "response_count": 8,
    "rows": {
        "1": "Peter Jackson",
        "2": "George Lucas",
        "3": "J.J. Abrams",
        "4": "Gareth Edwards",
        "5": "Irvin Kershner",
        "6": "Richard Marquand",
        "7": "Rian Johnson",
        "8": "Ron Howard"
    }
}
```

If there were lots of values, it would return something like this:

```json
{
    "total_count":  5000,
    "response_count": 100,
    "rows": {
        "1": "Peter Jackson",
        "2": "George Lucas",
        "3": "J.J. Abrams",
        "4": "Gareth Edwards",
        "5": "Irvin Kershner",
        "6": "Richard Marquand",
        "7": "Rian Johnson",
        ...
        "100": "Quentin Tarantino",
    }
}
```
By default, a maximum of 100 results will be returned at once. If `response_count < total_count` we should show the search UI rather than just a dropdown.

The ids endpoint should accept a search parameter, which can be used to filter what's returned. For example:

```json
// /api/tables/director/ids?search=george
{
    "total_count":  2,
    "response_count": 2,
    "rows": {
        "1": "George Lucas",
        "50": "George Armitage",
    }
}
```
The search UI could look something like this:

[...]

PR #34, opened 2021-02-16
fix clear filters for duration widget

sinisaos:
Small fix for clear filters in `RowFilter.vue`. Duration widget are not properly cleaned. Now all fields is properly cleaned.

### DIFF
--- a/admin_ui/src/components/RowFilter.vue
+++ b/admin_ui/src/components/RowFilter.vue
@@ -80,10 +80,32 @@ export default Vue.extend({
         },
         async clearFilters() {
             console.log("Clearing ...")
-            let form: HTMLFormElement = this.$refs.form
+            let form: any = this.$refs.form
+
+            let elements = form.elements
             form.reset()
+
+            for (let i = 0; i < elements.length; i++) {
+                let field_type = elements[i].type
+
+                switch (field_type) {
+                    case "text":
+                    case "textarea":
+                    case "hidden":
+                        elements[i].value = ""
+                        break
+                    case "select-one":
+                        elements[i].selectedIndex = 0
+                        break
+
+                    default:
+                        break
+                }
+            }
+
             this.$store.commit("updateFilterParams", {})
             this.$store.commit("updateCurrentPageNumber", 1)
+
             try {
                 await this.$store.dispatch("fetchRows")
             } catch (error) {
--- a/admin_ui/src/components/RowFilter.vue
+++ b/admin_ui/src/components/RowFilter.vue
@@ -81,27 +81,13 @@ export default Vue.extend({
         async clearFilters() {
             console.log("Clearing ...")
             let form: any = this.$refs.form
-
-            let elements = form.elements
-            form.reset()
-
-            for (let i = 0; i < elements.length; i++) {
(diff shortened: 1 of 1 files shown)


dantownsend:
@sinisaos Thanks for this.

I didn't realise that when you do `form.clear()`, it doesn't work on hidden inputs for some reason, and that's what the `DurationWidget` uses.

https://stackoverflow.com/questions/2559616/javascript-true-form-reset-for-hidden-fields

In your code you explicitly reset a few different field types in the for-loop - do you think we need to, or is it enough to just reset the hidden ones, as `form.reset` should've done it for the others?

Maybe we can just do this:

```javascript
[...form.elements].forEach((element) => {
    if (element.type == 'hidden') {
        element.value = ""
    }
})
```

sinisaos:
@dantownsend Sorry for the late reply. You are absolutely right. I'm making changes with the second commit. Is it ok now?

dantownsend:
@sinisaos Looks great, thanks!

It was such a subtle bug, I would never have found it if you hadn't noticed it.

sinisaos:
@dantownsend I accidentally came on this problem. 
I think also see little bug with Flatpickr in edit form on [small screen](https://flatpickr.js.org/mobile-support/). When open edit form datetime field is empty (which led to error on form submit) and I think that we need to show Flatpickr widget also on mobile screen (not browser native date/time/datetime input like is now). 
We need changes something like this.
 ```
InputField.vue

<flat-pickr
 v-bind:config="{ enableTime: true, disableMobile: true }"
    v-bind:name="getFieldName(title)"
    v-model="localValue"
></flat-pickr>
```
What do you think?

dantownsend:
@sinisaos Yes, you're right - I can replicate the problem on my phone.

Have you tried it with Flatpickr on mobile? Does it fit on the screen?

sinisaos:
@dantownsend No. I tried on Chromium developer tools mobile devices screens. On this it's fit.

mobile

dantownsend:
@sinisaos I tried it on the iOS simulator, for one of the smaller iPhones, and it fits OK.

Screenshot 2021-02-17 at 18 54 08

The UI feels a bit weird on iOS though. It seems fine on Android. It's better than not working though I guess.

I had a look for some alternative date time pickers, but couldn't find a good alternative. A lot of them only work on dates.

sinisaos:
@dantownsend I agree with you. I also think we need to make this change because it’s better to have a working form with a slightly worse look (it's not that bad) than a form that doesn’t work well. Are you going to make a change or me?

dantownsend:
@sinisaos I've just pushed the change - will release it shortly.

sinisaos:
@dantownsend Great. I am also making some changes to the foreign key selection from issue [#30](https://github.com/piccolo-orm/piccolo_admin/issues/30). It's not the best quality code, but it's work. It is a search autocomplete without changes in the backend endpoint and without the use of `HTML select`. The biggest problem is that I can't get FK readable (`1` is displayed instead of `Peter Jackson`) in edit form. Tommorow I can paste code in issue comment that you can see what it is about. Cheers.

dantownsend:
@sinisaos Sounds great, thanks.

PR #36, opened 2021-02-22
Improve foreign key selector

sinisaos:
Attempt to improve foreign key selector when dealing with lots of rows [#30](https://github.com/piccolo-orm/piccolo_admin/issues/30). Biggest issue is display foreign key readable in edit form. I try with `v-bind:value="$route.params.rowID ? row[keyName + '_readable']: ''"` in `RowForm.vue` with adding `?__readable=true` to `fetchSingleRow` in `store.ts`, but then I can't get FK primary key in hidden text input.


admin_ui/src/components/RowForm.vue
```diff
@@ -75,68 +72,35 @@ export default Vue.extend({
...
-            console.log("Received message")
-            app.baseIndex = app.baseIndex + 1
-        }
+        },
     },
-    destroyed() {
-        if (this.windowListener) {
-            window.removeEventListener("message", this.windowListener)
-        }
-    }
```
dantownsend, line 139: This stuff seems quite cryptic, but when you click on the edit or add button above the dropdown, it opens a new tab, and this code will tell the `KeySelect` element to refresh (and hence pull down fresh data from the API, containing the new / modified rows).

You'll see the bit of code which `opener.postMessage("edited row", document.location.origin)` in `EditRowForm`.

dantownsend:
@sinisaos Thanks for this.

It might be worth keeping the original component, and moving the new code into a component called something like `KeySearch`. Both are useful in different situations.

dantownsend:
One nice thing about using a drop down is handling null values is quite easy - it's just another value in the dropdown. I'm not sure how best to handle 'null' in a search box - just keeping it empty?

admin_ui/src/components/RowForm.vue
sinisaos: @dantownsend Thanks for the insight and explanation, it's clear now. I deleted this code because the component was refreshing and I had to select the value from the search dropdown twice.

sinisaos:
@dantownsend Honestly I forgot about null value. I don't think this attempt is a good one. If you think you can use something from this, that’s fine, but otherwise feel free to ignore this pull request.

sinisaos:
@dantownsend Or maybe use `KeySelect.vue` for adding and editing and use `KeySearch.vue` for `RowFilter.vue`?

admin_ui/src/components/RowForm.vue
dantownsend: It's a very obscure API which lets different tabs communicate with each other.

It does create weird issues though - sometimes when Chrome dev tools is open it sends loads of messages for some reason.

dantownsend:
@sinisaos This is a great start - but it's a surprisingly difficult thing.

sinisaos:
@dantownsend I made another commit where I use original component `KeySelect.vue` for adding and editing records (handling null values, FK readable etc.) and I use new component `KeySearch.vue` for filtering and searching (null value dosen't metter because already return 422 error and user can choose all records by leaving search bar empty or by clearing filters). Like you previously wrote I also think using two components for different situations is much better option. What do you think about this?

sinisaos:
@dantownsend Any feedback about this? If this is not good, you can ignore and close the pull request. Cheers.

dantownsend:
@sinisaos Yeah, sorry for the slow replies, I've been really busy in work.

I've been playing around with it. I'd like to integrate it with a search API, which I'm currently working on.

sinisaos:
@dantownsend No problem. I was just curious how we stand with this.

PR #37, opened 2021-03-07
added `inflate` option to example.py, for generating lots of dummy data

dantownsend:
Added an option to the `example.py` CLI, which will populate lots of extra data in the database.

```
# This will add 10,000 extra rows of data
python -m piccolo_admin.example --inflate=10000
```

This is really useful when testing the performance of the admin with large data sets.


dantownsend:
Will be used for testing https://github.com/piccolo-orm/piccolo_admin/pull/36

PR #39, opened 2021-03-10
Foreign key selector

dantownsend:
Building on top of https://github.com/piccolo-orm/piccolo_admin/pull/36

The ids endpoint now supports search, so modified accordingly.


dantownsend:
@sinisaos I think I'm finally finished with this now.

I didn't make many changes to your implementation - just added search to the IDs endpoint. It took ages because implementing search on the API was quite tricky. And I needed a way to test it with lots of data, so added that.

What do you think? Is there anything I've missed before merging it in?

sinisaos:
@dantownsend I tried the changes yesterday and I think it looks and works good. Cheers.

dantownsend:
@sinisaos Thanks - will merge it in later on.

PR #41, opened 2021-03-16
additional FK improvements

sinisaos:
Using `KeySearch.vue` for adding and editing records, because select dosen't work well [with large database tables](https://piccolo-orm.com/blog/building-an-admin-to-handle-millions-of-rows/)


sinisaos:
@dantownsend I did this PR because select doesn’t work well with large datasets. The search API works well on adding rows and editing rows when the FK input field change by user. The problem is when the FK input field doesn't change by user, because then error 422 is returned because the readable presentation of FK is not integer. You can try and you will see what I mean. Do you have any idea how to trigger FK id  (which is on `KeySearch.vue` hidden field) on form submit when FK input field is not changed? Cheers.

dantownsend:
@sinisaos Thanks for this.

I can see the issue. I think it's almost there - the `KeySearch` just needs to accept two values, `id` and `readable`, instead of just a single `value`.

sinisaos:
@dantownsend Yes, exactly. But I don't know how to achieve that. I only can get both fields to be either `id` (with `v-bind:value="getValue(property.title)"`) or` readable` (with `v-bind:value="$route.params.rowID ? row[keyName + '_readable']: ''"`). Maybe you know how to do that?

dantownsend:
@sinisaos I've pushed some changes to your pull request - does that fix it?

sinisaos:
@dantownsend That's great. Thank you. Now everything works great :).

dantownsend:
@sinisaos Cool - will merge it in. Thanks again.

sinisaos:
@dantownsend I tried a table with multiple FK's and it also works great, and also table with 100000 rows and search api is much better than select and work seamlessly and instantly. Thank you. Cheers.

dantownsend:
@sinisaos Yeah, it's much better now. I tried the old UI with 100,000 rows, and it crashed the Chrome tab!

sinisaos:
@dantownsend I also had a browser crash. Now is great.

PR #49, opened 2021-04-14
Small changes to login page

sinisaos:
Small changes to login page to include `site_name` in template.


dantownsend:
It's a nice idea.

The only problem was the meta endpoint required login, so it couldn't always fetch the site name. I've made this available without auth, and it seems to work.

sinisaos:
@dantownsend Great. Thank you. I forgot about that. Your changes are working and everything is fine now.

PR #60, opened 2021-05-27
small docs fix

sinisaos:
Small docs fix. I think you mean `choices` attribute set. If I'm wrong, just close this PR. Cheers.


dantownsend:
@sinisaos Oops - good catch!

PR #68, opened 2021-06-27
Add custom Swagger UI from piccolo_api

dantownsend:
The Swagger docs in the Piccolo admin should now work with POST requests (Piccolo's CSRF middleware used to block it).

### DIFF
--- a/piccolo_admin/endpoints.py
+++ b/piccolo_admin/endpoints.py
@@ -11,6 +11,7 @@ from piccolo.apps.user.tables import BaseUser
 from piccolo_api.csrf.middleware import CSRFMiddleware
 from piccolo_api.crud.endpoints import PiccoloCRUD
 from piccolo_api.fastapi.endpoints import FastAPIWrapper, FastAPIKwargs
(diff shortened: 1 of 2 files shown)


dantownsend:
@sinisaos I think I finally worked out how to get POST requests working in the Swagger UI, with the CSRF middleware.

In the Swagger UI there's a `requestInterceptor` parameter, which can be used to add the CSRF token.

https://github.com/piccolo-orm/piccolo_api/blob/3c5ea4d811ef35d7f008b305891c310c7228db86/piccolo_api/openapi/templates/swagger_ui.html.jinja#L33

sinisaos:
@dantownsend That's great. I tried a new feature and everything works great. Maybe you can change [hint](https://piccolo-admin.readthedocs.io/en/latest/api_documentation/index.html) to something like `From version 0.14.0 the Swagger interactive features work with all HTTP methods` because as of version 0.14.0 there is no such restriction anymore. Or you can completely remove that hint from docs.

dantownsend:
@sinisaos That's a good point - I forgot to update the docs to reflect the change, thanks!

PR #75, opened 2021-07-21
add scripts folder

### LINKED ISSUE #72: Add scripts folder
Following [this](https://github.com/piccolo-orm/piccolo/pull/133):

- Move all the scripts to the `scripts` folder
- Update github workflows to use the scripts

aminalaee:
Closes #72 
- Added the release.yaml, you need to get PyPI API token and set up `PYPI_TOKEN` secret for the project or organization in the Settings -> Secret.
- The results are not uploading to codecov yet. You need to set `CODECOV_TOKEN` secret for the project or organization
- Added isort (with black profile) and black to lint check to keep the codebase consistent.

If you could please enforce `Squash and merging` for all `piccolo` projects. I realized all my 20 commits in a PR were merged into master. That would pollute the history a bit. We can squash and merge PR with one commit into master, if you agree. If we decide to enforce it in the settings, we won't  forget to do it :)
I think that can be set for the whole organization.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+per-file-ignores = __init__.py:F401
+max-line-length = 79
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,24 @@
+name: Release
+
+on:
+  release:
+    types:
+      - created
+
+jobs:
+  release:
+    name: "Publish release"
+    runs-on: "ubuntu-latest"
+
+    steps:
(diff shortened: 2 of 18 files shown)


dantownsend:
@aminalaee Yeah, I've always felt a bit uneasy about squashing commits, as it occasionally causes issues.

This is quite a nice article about it:

https://sdtimes.com/bugs/guest-view-hidden-hazards-squashing-git-commits/

You're right though, it does create a cleaner history, so does make sense in most situations.

Thanks again for making these changes.

Issue #82, opened 2021-08-09
Cypress tests


dantownsend:
When new features are added to Piccolo, a lot of manual testing is involved with Piccolo Admin.

It would be much nicer to have a tool like [Cypress](https://www.cypress.io/), and a simple suite of automated tests.

Initially, having the following would be great:

 * Testing that the login page works
 * Testing that the list page works, with some basic filtering
 * Testing that the detail page works (ideally with adding / editing a row)

dantownsend:
First version Implemented in https://github.com/piccolo-orm/piccolo_admin/pull/84

PR #84, opened 2021-08-12
add cypress tests

sinisaos:
Based on #82


dantownsend:
@sinisaos This is really cool, thanks.

I only know the basics about Cypress, but it seems like a valuable tool. Will have a play around with what you've added.

dantownsend:
@sinisaos Wow, this is awesome. Just testing it now.

Screenshot 2021-08-13 at 21 49 19

dantownsend:
@sinisaos I'll merge this in. I have a few ideas for other Cypress related things we can try, but no need to overly bloat this ticket.

Issue #87, opened 2021-08-14
Upgrade to Vue 3


dantownsend:
Vue 2 should still be fine for some time, but Vue 3 is definitely a step up. A first step is working out how much work is involved.

sinisaos:
@dantownsend I started playing with this. There are a lot of changes, but with [this help](https://v3.vuejs.org/guide/migration/migration-build.html#overview) I managed to log in and logged out to admin, but for now the row listing doesn't work. I think the biggest problem will be with libraries compatibility (e.g. `vue-json-excel` for CSV export doesn't work with Vue3 and I don't see much activity in their repository). If you want I can do PR so we can track changes and progress.

dantownsend:
@sinisaos If you could create a PR that would be amazing. The upgrade definitely isn't painless, but the Typescript improvements are worthwhile.

sinisaos:
@dantownsend Great. I will do PR that you can see changes in code.

dantownsend:
Completed in Piccolo Admin v1

PR #89, opened 2021-08-16
initial commit for migrate from Vue2 to Vue3

sinisaos:
Based on #87.
Work in progress
- [x] fixing login process
- [x] replacing deprecated features form Vue2 (Vue.set(), Vue.delete(), filters in template) with Vue3 features
- [x] upgraded major library dependencies for working with Vue 3
- [x] removing vue-json-excel and making our own implementation of CSV export
- [x] fixing RowListing and CRUD
- [x] fixing sorting and filtering
- [x] fixing reference table search results
- [x] added basic types
- [x] custom forms
- [x] fixing nullable fields
- [x] TableConfig


sinisaos:
@dantownsend I see a lot of activity in the main Piccolo ORM repo, but when you have time just try the changes I’ve made so far and I’m interested in your opinion. The main problem is that `rows` is not updated when `currentTableName` changed. This is probably because the api has changed in Vuex 4. I have yet to investigate this. Cheers.

dantownsend:
@sinisaos I agree, the Vuex changes are problematic. I need to read the docs fully to understand the complete impact.

The only thing which jumped out at me so far, is we've removed `Vue.extend` and are returning plain objects.

```javascript
return {
    data() {}
    ...
}
```

Even though this works, Vue 3 added something called `defineComponent`. 

```javascript
return defineComponent({
    data() {}
    ...
})
```

If we use this instead of a plain object, it might help with Typescript support.

dantownsend:
I've had a look at the Vuex migration guide:

https://next.vuex.vuejs.org/guide/migrating-to-4-0-from-3-x.html

It's not quite as bad as I feared. Do you want me to push the Vuex changes?

sinisaos:
@dantownsend I push new commit and now crud, sorting and filtering work. I'm not Vue and Typescript expert and any help is welcome.

dantownsend:
@sinisaos Great, thanks. I'll pull it down, and see where I can help out.

sinisaos:
@dantownsend Great, thanks. You are right about [defineComponent](https://v3.vuejs.org/guide/typescript-support.html#defining-vue-components).

sinisaos:
@dantownsend I think I have completed the changes to switch to Vue 3. Everything seems to be working ok. Can you try this and I want to hear your opinion and suggestions. Based on reading the articles online, I think we have to wait for the transition to Vue 3 because the ecosystem is not mature enough at the moment.

admin_ui/src/App.vue
```diff
@@ -50,17 +54,11 @@ export default Vue.extend({
...
-                                    nextURL: nextURL,
-                                },
-                            })
-                        }, 0)
-                    }
+                    setTimeout(function () {
+                        app.$router.push({
+                            name: "login",
+                        })
+                    }, 0)
```
dantownsend, line 61: What was the issue here? Are we unable to access `app.$route.path`?

admin_ui/src/components/ArrayWidget.vue
```diff
@@ -44,20 +54,28 @@ export default {
...
     },
     methods: {
-        updateArray($event, index) {
-            this.$set(this.internalArray, index, $event.target.value)
+        updateArray($event: Event, index: number) {
+            Reflect.set(
+                this.internalArray,
+                index,
+                ($event.target as HTMLInputElement).value
+            )
```
dantownsend, line 66: I didn't know about `Reflect.set`, but looks good!

admin_ui/src/components/CSVButton.vue
```diff
@@ -59,21 +54,27 @@ export default {
...
+                    ),
+                ]
+                    .join("\n")
+                    .replace(/(^\[)|(\]$)/gm, "")
+                const data: string = encodeURI(csv)
+                const link: HTMLAnchorElement = document.createElement("a")
+                link.setAttribute("href", data)
+                link.setAttribute("download", `${this.tableName}.csv`)
+                link.click()
+                alert("Your data is ready.")
```
dantownsend, line 71: It looks like you were able to replicate the download to CSV without using the library. Is this correct?

admin_ui/src/components/DurationWidget.vue
```diff
@@ -88,11 +90,11 @@ export default {
...
-            weeks: undefined,
-            days: undefined,
-            hours: undefined,
-            minutes: undefined,
-            seconds: undefined,
+            weeks: undefined as any,
+            days: undefined as any,
+            hours: undefined as any,
+            minutes: undefined as any,
+            seconds: undefined as any,
```
dantownsend, line 97: Instead of `undefined as any`, does `undefined as Number | undefined` work?

admin_ui/src/components/EditRowForm.vue
```diff
@@ -1,51 +1,37 @@
...
-                        <li>
-                            <DeleteButton
-                                :includeTitle="true"
-                                class="subtle"
-                                v-on:triggered="deleteRow"
-                            />
-                        </li>
-                    </DropDownMenu>
-                </a>
-            </p>
```
dantownsend, line 23: Did we lose this because it's possible to delete from the row list page?

admin_ui/src/components/EditRowForm.vue
```diff
@@ -60,31 +46,31 @@ export default Vue.extend({
         }
     },
     computed: {
-        schema() {
+        schema(): any {
             return this.$store.state.schema
         },
-        selectedRow() {
+        selectedRow(): any {
```
dantownsend, line 52: We need to try and avoid `any` is possible. Does `object` work here?

admin_ui/src/components/InputField.vue
```diff
@@ -194,14 +198,14 @@ export default Vue.extend({
             return name.toLowerCase().split(" ").join("_")
         },
         setTextareaHeight() {
-            let element = this.$refs.textarea
+            let element: any = this.$refs.textarea
```
dantownsend, line 201: We might be able to use HTMLElement as the type here

admin_ui/src/components/KeySearch.vue
```diff
@@ -42,18 +40,19 @@
...
 import { FetchIdsConfig } from "../interfaces"
 
-export default {
+export default defineComponent({
     props: {
         fieldName: String,
         tableName: String,
-        rowID: undefined,
-        readable: undefined,
+        rowID: undefined as any,
```
dantownsend, line 50: `number | undefined` might work here

admin_ui/src/components/KeySearch.vue
```diff
@@ -42,18 +40,19 @@
...
+        readable: undefined as any,
```
dantownsend, line 51: `boolean | undefined` might work here

admin_ui/src/components/KeySearch.vue
```diff
@@ -42,18 +40,19 @@
...
     },
     data() {
         return {
-            ids: [],
+            ids: [] as any,
```
dantownsend, line 55: `number[]` might work here

admin_ui/src/components/KeySearch.vue
```diff
@@ -79,7 +78,7 @@ export default {
                 return 0
             })
         },
-        selectResult(id, readable) {
+        selectResult(id: undefined, readable: undefined) {
```
dantownsend, line 81: The type here looks a bit odd.

I'd try `selectResult(id: number | undefined, readable: boolean | undefined)`

admin_ui/src/components/ReferencingTables.vue
```diff
@@ -55,28 +54,32 @@ export default Vue.extend({
...
-                query
+                query,
             }
+            // updated store to get correct reference table search results
+            this.$store.commit("updateFilterParams", query)
             let vueUrl = this.$router.resolve(location).href
             window.open(
                 `${document.location.origin}${document.location.pathname}${vueUrl}`,
-                "_blank"
+                "_self"
```
dantownsend, line 76: We might be worth keeping this as "_blank" so it opens in a new tab. Otherwise if they have unsaved form data, it might get lost.

admin_ui/src/components/RowFilter.vue
```diff
@@ -16,22 +28,22 @@
...
     components: {
         RowFormSearch,
     },
     computed: {
-        schema() {
+        schema(): any {
             return this.$store.state.schema
         },
-        tableName() {
+        tableName(): undefined {
```
dantownsend, line 46: `string | undefined`

admin_ui/src/components/RowFormSearch.vue
```diff
@@ -59,23 +66,21 @@
...
+import { defineComponent } from "vue"
 import KeySearch from "./KeySearch.vue"
 import InputField from "./InputField.vue"
 
-export default Vue.extend({
+export default defineComponent({
     props: {
         row: Object,
-        schema: Object,
+        schema: Object as any,
```
dantownsend, line 83: Vue 3 has a thing called `PropType`. Here we might want to try:

```
schema: Object as PropType<object>
```

admin_ui/src/components/RowFormSelect.vue
```diff
@@ -24,40 +27,29 @@
...
-                                tableName: property.extra.to,
-                                rowID: getKeySelectID(property.title),
-                            },
-                        }"
-                        class="add"
-                        target="_blank"
-                        v-if="!isFilter"
-                    >
-                        <font-awesome-icon icon="edit" />
-                    </router-link>
```
dantownsend, line 41: Do we definitely want to remove this?

dantownsend:
@sinisaos You've done a great job here, thanks.

In terms of transitioning to Vue 3, what are the main things which are missing in your opinion?

admin_ui/src/App.vue
sinisaos: There is no issue but new way does the same thing with much cleaner url (`http://piccolo_admin/#/login/` vs `http://piccolo_admin/#/login/?nextURL=/login/`)

admin_ui/src/components/ArrayWidget.vue
sinisaos: You are right. [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) is new global object introduce in ES6 that allow handle dynamic code.

admin_ui/src/components/CSVButton.vue
sinisaos: Yes. It does the same thing without an external library because `vue-json-excel` doesn't have support for Vue 3 and based on repo activities there probably won't be any support soon.

admin_ui/src/components/DurationWidget.vue
sinisaos: No. If we don't want to use `undefined as any` only setting default to 0 works
```javascript
weeks: 0,
days: 0,
hours: 0,
minutes: 0,
seconds: 0,
```
I can change this if you think it's better.

admin_ui/src/components/EditRowForm.vue
sinisaos: We already have possibility to delete single row from the row list page and I remove delete from edit page because when delete row I get `404` for deleted row id (`Vuex` doesn't update state of all rows). It was easier to me to remove option to delete row from edit.

admin_ui/src/components/EditRowForm.vue
sinisaos: I agree with you that we need to try and avoid `any` if is possible but unfortunately `object` does not work.

admin_ui/src/components/InputField.vue
sinisaos: Unfortunately `HTMLElement` or `HTMLTextAreaElement` does not work.

admin_ui/src/components/KeySearch.vue
sinisaos: Unfortunately `number | undefined` does not work

admin_ui/src/components/KeySearch.vue
sinisaos: Unfortunately `boolean | undefined` does not work

admin_ui/src/components/KeySearch.vue
sinisaos: Unfortunately `number[]` does not work. `[] as number[]` or `Array<number>` also does not work

admin_ui/src/components/KeySearch.vue
sinisaos: Unfortunately `selectResult(id: number | undefined, readable: boolean | undefined)` does not work. Only `selectResult(id: any | undefined, readable: any | undefined)` work.

admin_ui/src/components/RowFilter.vue
sinisaos: `tableName(): string | undefined` works great.

admin_ui/src/components/RowFormSearch.vue
sinisaos: Unfortunately does not work. I tried `schema: Object as PropType<object>` but then `this.schema?.required` error `Property 'required' does not exist on type 'object'`.

admin_ui/src/components/RowFormSelect.vue
sinisaos: No, we keep this. I probably accidentally deleted this.

admin_ui/src/components/ReferencingTables.vue
sinisaos: I agree with you that `_blank` is better solution, but search result i that case is wrong. Only with opening in same window with `_self` I get correct reference table search results adding [this](https://github.com/sinisaos/piccolo_admin/blob/upgrade_to_Vue3/admin_ui/src/components/ReferencingTables.vue#L71)

sinisaos:
@dantownsend Thanks. 
In terms of switching to Vue 3 based on what I’ve read online I think the compatibility of external libraries is a major issue. I don't think that's the biggest problem in our case because we don't use a lot of external libraries (the ones we use have been updated for Vue 3 and we've replaced CSV export to not use `vue-json-excel`).
What bothers me the most is when I try `Piccolo Admin` with more tables  and more complex foreignkeys sometimes the rows doesn't update properly and the readable FK columns doesn't display well. I don't understand but sometimes everything is ok and sometimes  doesn't. Whether it’s a `Vuex` problem or something else. I don’t know.

admin_ui/src/App.vue
dantownsend: The only advantage with the old approach, is let's say you're on a page and you go away and make a cup of tea. You come back, and your session has expired. When you login, it will take you back to the page you were previously on.

admin_ui/src/components/DurationWidget.vue
dantownsend: Yes, good idea - I think 0 is cleaner.

admin_ui/src/components/KeySearch.vue
dantownsend: OK, let's just have it as `any` for now then. I think having `undefined` is a strange looking one, as if the argument was always of type undefined, it wouldn't be a very useful function!

admin_ui/src/components/ReferencingTables.vue
dantownsend: Ah, I see. Better to keep what you've done then.

admin_ui/src/components/RowFormSearch.vue
dantownsend: OK, no problem. With a lot of these type issues we'll be able to solve them in the future by using the composition API more.

dantownsend:
@sinisaos Yeah, we need to be confident that the app will still work nicely before releasing it.

This is why I'm so excited by the Cypress stuff that you did, as it gives us more confidence that the app is working as expected.

I was watching this video to try and work out how we could automatically run the Cypress tests in GitHub actions:

https://www.youtube.com/watch?v=gokM_zEmWLA

I've been quite busy fixing issues on the main Piccolo repo, but I want to focus on the admin again, as it's such an important part of the ecosystem.

sinisaos:
@dantownsend No hurry. I'll be pushed commits and when you have time you just pull branch, try it and write your suggestions and opinion. We still have things to fix.

admin_ui/src/App.vue
sinisaos: You are right. I moved back to the old approach.

admin_ui/src/components/ReferencingTables.vue
sinisaos: I managed to use `_blank` and open in new window because is better solution, but I must remove edit option from `RowFormSelect` due to error of `rowID` is `undefined`.

admin_ui/src/components/RowFormSelect.vue
sinisaos: For now I remove this because  to of `rowID` is undefined in `ReferencingTables.vue`. I think that problem is [here](https://github.com/piccolo-orm/piccolo_admin/blob/master/admin_ui/src/components/RowFormSelect.vue#L118). I tried `Reflect.set()` and `this.keySelectIDs[propertyTitle] = value` but neither way works.

admin_ui/src/components/EditRowForm.vue
sinisaos: This feature returns with commit [7364542](https://github.com/piccolo-orm/piccolo_admin/pull/89/commits/736454244de6cfc006d16d766ab016a61b9613f2).

admin_ui/src/components/RowFormSelect.vue
sinisaos: This feature returns with commit [7364542](https://github.com/piccolo-orm/piccolo_admin/pull/89/commits/736454244de6cfc006d16d766ab016a61b9613f2).

sinisaos:
@dantownsend I think I have completed the transition. All features remain the same. Can you please review it?

dantownsend:
@sinisaos Thanks. I'll start reviewing it.

I might not finish it today though - I'll pull it down and test it out thoroughly.

admin_ui/src/components/EditRowForm.vue
```diff
@@ -133,17 +138,6 @@ export default Vue.extend({
...
-        "$route.params.tableName": async function () {
-            await Promise.all([
-                this.fetchData(),
-                this.$store.dispatch("fetchSchema", this.tableName),
-            ])
-        },
-        "$route.params.rowID": async function () {
-            await this.fetchData()
-        },
-    },
```
dantownsend, line 146: Did this not work anymore? I think the original intention here was if someone hand modified the URL then the page would update appropriately.

admin_ui/src/components/RowFormSelect.vue
```diff
@@ -109,13 +113,8 @@ export default Vue.extend({
...
         getKeySelectID(propertyTitle: string) {
-            return (
-                this.keySelectIDs[propertyTitle] || this.getValue(propertyTitle)
-            )
-        },
-        keySelectChanged(propertyTitle: string, value: number) {
-            console.log(`${propertyTitle} = ${value}`)
-            Vue.set(this.keySelectIDs, propertyTitle, value)
+            localStorage.setItem("rowID", this.getValue(propertyTitle))
+            return localStorage.getItem("rowID")
```
dantownsend, line 117: This might not have worked because `Vue.set` was removed in Vue 3. If we replace it with `Reflect.set`, or just `this.keySelectIDs[propertyTitle] = value` does it work as it was before?

admin_ui/src/components/EditRowForm.vue
sinisaos: No. I had to remove all the route watchers because things are not working properly with them.

admin_ui/src/components/RowFormSelect.vue
sinisaos: I’ve tried both before, but neither way works. I think localStorage is the good way to store FK `rowID` via browser tabs, and since `rowID` is not sensitive information, I think that's fine.

admin_ui/src/components/EditRowForm.vue
dantownsend: What a pain. It might be possible with the composition API, but I can imagine a lot of people's code being broken by removing `$route`.

sinisaos: You are actually right. I returned this piece of code in the last commit because the problem was actually in the FK `rowID` and I deleted this code before solving that issue because I didn't know what the problem was. Thanks for your suggestion.

sinisaos:
Closed - open for too long without rejection, merge or activity. Feel free to reopen PR and use it again if code is still relevant.

Issue #69, opened 2021-06-27
Increase test coverage for `create_admin`


dantownsend:
Just a reminder to myself to increase code coverage for `create_admin`.

sinisaos:
@dantownsend I have a question. If we omit files which only usage is to serve demo app for admin in` pyproject.toml` like this
```
[tool.coverage.run]
omit = ["piccolo_admin/example.py", 
        "piccolo_admin/example_data.py",
        "piccolo_admin/sandbox.py"] 
```
 we can easily achieve coverage over 90. Is this ok or we need to test all files?

dantownsend:
@sinisaos You're right, those files aren't particularly important, and can be omitted from code coverage.

The tests in general for Piccolo Admin are very light, considering how much it does. A lot of it is covered in Piccolo API, but there are some methods in `AdminRouter` which are custom to Piccolo Admin, and I don't think have any coverage:

 * get_root
 * get_user
 * get_meta
 * get_table_list
 * get_site_name

It would be nice to have simple tests for these.

sinisaos:
@dantownsend  `get_root` and `get_meta` are not problem because is not protected by authentication. Something like this.
```python
def test_get_meta(self):
    client = TestClient(APP)

    response = client.get("/meta/")
    self.assertEqual(
        response.json(),
        {"piccolo_admin_version": "0.14.0", "site_name": "Piccolo Admin"},
    )
    self.assertEqual(response.status_code, 200)

def test_get_root(self):
    client = TestClient(APP)

    response = client.get("/")
    self.assertEqual(response.status_code, 200)
```

We even don't need this function `get_site_name` because we can get site_name from `get_meta`.

But, the `get_user` and `get_table_list` functions are protected by authentication and testing with client.post() doesn't work because the CSRF token is missing. The best I can do is check the auth failed.
```python
def test_get_tables_list(self):
    client = TestClient(APP)

    response = client.get("/api/tables/")
    self.assertEqual(response.json(), {"error": "Auth failed"})
    self.assertEqual(response.status_code, 401)
```

sinisaos:
@dantownsend I think you can close this issue because it’s done.

PR #90, opened 2021-08-16
increase coverage

sinisaos:
Based on #69

### DIFF
--- a/piccolo_admin/endpoints.py
+++ b/piccolo_admin/endpoints.py
@@ -10,6 +10,7 @@ from functools import partial
 
 from fastapi import FastAPI
 from piccolo.apps.user.tables import BaseUser
+from piccolo.columns.reference import LazyTableReference
 from piccolo_api.crud.endpoints import PiccoloCRUD
 from piccolo_api.csrf.middleware import CSRFMiddleware
 from piccolo_api.fastapi.endpoints import FastAPIKwargs, FastAPIWrapper
@@ -200,8 +201,7 @@ class AdminRouter(FastAPI):
 
     def get_user(self, request: Request) -> UserResponseModel:
         return UserResponseModel(
-            username=request.user.display_name,
-            user_id=request.user.user_id,
+            username=request.user.display_name, user_id=request.user.user_id,
         )
 
     ###########################################################################
@@ -237,9 +237,15 @@ def get_all_tables(
             for i in table._meta.foreign_key_columns
         ]
         for reference in references:
-            if reference not in output:
-                output.append(reference)
-                get_references(reference)
+            table = (
+                reference.resolve()
+                if isinstance(reference, LazyTableReference)
+                else reference
+            )
+
+            if table not in output:
+                output.append(table)
+                get_references(table)
 
     for table in tables:
         if table not in output:
--- a/piccolo_admin/endpoints.py
+++ b/piccolo_admin/endpoints.py
@@ -107,7 +107,7 @@ class AdminRouter(FastAPI):
 
         api_app.add_api_route(
             path="/tables/",
-            endpoint=self.get_table_list,
+            endpoint=self.get_table_list,  # type: ignore
             methods=["GET"],
             response_model=t.List[str],
             tags=["Tables"],
@@ -115,7 +115,7 @@ class AdminRouter(FastAPI):
 
         api_app.add_api_route(
             path="/meta/",
-            endpoint=self.get_meta,
+            endpoint=self.get_meta,  # type: ignore
             methods=["GET"],
             tags=["Meta"],
             response_model=MetaResponseModel,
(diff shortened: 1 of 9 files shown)


sinisaos:
@dantownsend @aminalaee Sorry for interrupt but I need help. What is right way or syntax to omit certain files and folders from coverage in pyproject.toml? I tried 
```
[tool.coverage.report]
omit = ["*/piccolo_app.py",
        "*/piccolo_migrations/*"] 
```
and
```
[tool.coverage.run]
omit = ["*/piccolo_app.py",
        */piccolo_migrations/*"]
```
but when I create PR, none of these methods work and tests checks failed. Locally everything works ok.

aminalaee:
@sinisaos I think it should be coverage.report

sinisaos:
@aminalaee Thanks. I’ve tried both, but neither way works. Strange, but I'm probably doing something wrong.

aminalaee:
@sinisaos I'll take a look tomorrow if it wasn't resolved.

sinisaos:
@aminalaee Thanks.

dantownsend:
@sinisaos It looks OK to me. From what I've read online, `tool.coverage.run` should work.

All I can think is:

 * Is there a version mismatch, i.e. are you running an older version on your local machine?
 * Are the path globs correct? Maybe this instead?

```python
[tool.coverage.run]
omit = ["**/piccolo_app.py",
        **/piccolo_migrations/*"]
```

sinisaos:
@dantownsend From codecov logs for py3.8 and py3.9 I must add `requests` in `test-requirements`. For py3.7 error is `Error: The operation was canceled`. I think is time to install py3.7 on my local machine and try. Probably syntax for `[tool.coverage.run]` is fine and that may not have been a problem at all. Thanks for your help and sorry for inconvenience.

sinisaos:
@dantownsend Sorry for this mess, but I finally made it. The problem was the lack of a dist folder in the `piccolo_admin` folder. I also need to update the github workflows for linters. Cheers.

piccolo_admin/endpoints.py
```diff
@@ -222,9 +222,6 @@ def get_table_list(self, request: Request) -> t.List[str]:
 
     ###########################################################################
 
-    def get_site_name(self, request: Request) -> JSONResponse:
-        return JSONResponse({"site_name": self.site_name})
```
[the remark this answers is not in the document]
dantownsend: Scrap that, doesn't seem like it's used - seems fine to delete it.

scripts/run-tests.sh
```diff
@@ -5,4 +5,11 @@
 # To run a single test tests/test_foo.py::TestFoo::test_foo
 
 export PICCOLO_CONF="tests.piccolo_conf"
-python -m pytest --cov=piccolo_admin --cov-report xml --cov-report html --cov-fail-under 23 -s $@
+cd admin_ui
+npm install
+npm run build
+rm -rf node_modules
```
dantownsend, line 11: You're right to do this, as Jinja freaks out if it can't find a template.

Building the front end is quite a time consuming process though. An alternative might be to create the dist folder, and put a dummy index file in there.

```
mkdir piccolo_admin/dist
touch piccolo_admin/dist/index.html
```

sinisaos: Great idea. I don't have time now but I'll try later

sinisaos: This works
```
mkdir piccolo_admin/dist
mkdir piccolo_admin/dist/css
mkdir piccolo_admin/dist/js
touch piccolo_admin/dist/index.html
python -m pytest --cov=piccolo_admin --cov-report xml --cov-report html --cov-fail-under 85 -s $@
rm -rf piccolo_admin/dist
```

dantownsend: Cool, will try and finish reviewing this today.

piccolo_admin/endpoints.py
```diff
@@ -107,23 +108,23 @@ def __init__(
 
         api_app.add_api_route(
             path="/tables/",
-            endpoint=self.get_table_list,
+            endpoint=self.get_table_list,  # type: ignore
```
dantownsend, line 111: I tried fixing this so we don't need `type: ignore` but seems tricky.

tests/test_endpoints.py
```diff
@@ -22,3 +24,28 @@ class TestGetAllTables(TestCase):
...
+
+
+class TestAdminRouter(TestCase):
+    def test_get_meta(self):
+        client = TestClient(APP)
+
+        response = client.get("/meta/")
+        self.assertEqual(
+            response.json(),
+            {"piccolo_admin_version": "0.14.0", "site_name": "Piccolo Admin"},
```
dantownsend, line 36: This will break in future versions of Piccolo admin.

We might have to do

```python
from piccolo_admin.version import __VERSION__

{"piccolo_admin_version": __VERSION__, "site_name": "Piccolo Admin"},
```

dantownsend: I'll push this change.

PR #93, opened 2021-09-21
nicer json  representation in UI

sinisaos:
Nicer json representation in admin UI


dantownsend:
Thanks for this.

It looks good in the listing view. My only concern is if the JSON is massive, we might have to truncate it.

Screenshot 2021-09-24 at 12 48 38

I can't get the nice formatting to display in the detail view:

Screenshot 2021-09-24 at 12 48 43

Any ideas?

sinisaos:
@dantownsend I don't know. In Firefox and Chromium looks good

FireShot Capture 038 - Piccolo Admin - piccolo_admin

I agree with you that we must truncate it.

dantownsend:
@sinisaos OK, let me try again.

sinisaos:
@dantownsend Also work in [Epiphany](https://en.wikipedia.org/wiki/GNOME_Web)  which is web browser based on the GTK port of Apple's WebKit rendering engine.

dantownsend:
@sinisaos I'm going to push one commit to this PR - just to register `Studio` with the example app, so it's easier to test JSON fields.

dantownsend:
@sinisaos It works now - I don't think I had the latest code pulled down when I tested it last time.

sinisaos:
@dantownsend And you also made a proxy to Pydantic serialization from main Piccolo repo to Piccolo API repo. Great.

sinisaos:
@dantownsend We can truncate json with existing `abbreviate` filter. If you want you can push commit with these if you like it?

```Vue
<span v-else-if="isJSON(name)">
    <pre>{{ row[name] | formatJSON | abbreviate }}</pre>
</span>
```
FireShot Capture 040 - Piccolo Admin - piccolo_admin

dantownsend:
@sinisaos Yeah, that looks good. Are you able to push the change? If not I'll do it.

admin_ui/src/components/MessagePopup.vue
```diff
@@ -1,7 +1,22 @@
 <template>
     <div
-        id="message_popup"
-        v-if="visible"
+        id="message_error"
+        v-if="visible && message.includes('error')"
```
dantownsend, line 4: Rather than checking if the message contains 'error', we should use the `apiResponseMessage`, which has a `type` attribute. This will be `error` if there's an error.

admin_ui/src/components/MessagePopup.vue
```diff
@@ -1,7 +1,22 @@
...
+                href="#"
+                v-on:click.prevent="visible = false"
+            >
+                <font-awesome-icon icon="times" />
+            </a>
+        </p>
+    </div>
+    <div
+        id="message_success"
+        v-else-if="visible"
```
dantownsend, line 19: Rather than duplicating the code, it might be cleaner to have a dynamic class or style applied which just sets the background red.

Something like:

```
:class="{error: apiResponseMessage.type == 'error'}"
```

sinisaos: Like this

```Vue
<template>
    <div
        id="message_popup"
        v-bind:style="{'background-color':'red'}"
        v-if="visible && apiResponseMessage.type == 'error'"
    >
        <p class="message">{{ message }}</p>
        <p class="close">
            <a
                class="subtle"
                href="#"
                v-on:click.prevent="visible = false"
            >
                <font-awesome-icon icon="times" />
            </a>
        </p>
    </div>
    <div
        id="message_popup"
        v-else-if="visible"
    >
        <p class="message">{{ message }}</p>
        <p class="close">
            <a
                class="subtle"
                href="#"
                v-on:click.prevent="visible = false"
            >
                <font-awesome-icon icon="times" />
            </a>
        </p>
    </div>
</template>


<script lang="ts">
import Vue from "vue"

export default Vue.extend({
    data() {
        return {
            visible: false,
            timeLastAppeared: 0,
        }
    },
    computed: {
        message(): string {
            return this.apiResponseMessage
                ? this.apiResponseMessage.contents
                : "-"
        },
        apiResponseMessage(): i.APIResponseMessage {
            return this.$store.state.apiResponseMessage
        },
    },
    methods: {
        getTime(): number {
            return new Date().getTime()
        },
    },
    watch: {
        visible() {
            this.timeLastAppeared = this.getTime()
        },
        apiResponseMessage() {
            this.visible = true
            const app = this

            setTimeout(() => {
                const now = this.getTime()
                // Only hide if it if it wasn't subsequently opened again.
                if (now - app.timeLastAppeared >= 3000) {
                    app.visible = false
                }
            }, 3000)
        },
    },
})
</script>


<style lang="less">
@import "../vars.less";

div#message_popup {
    display: flex;
    position: fixed;
    bottom: 0;
    width: 100%;
    padding: 0.5rem 1rem;
    box-sizing: border-box;
    background-color: green;

    p,
    a {
        color: white !important;
    }

    p.message {
        flex-grow: 1;
        text-align: center;
    }

    p.close {
        flex-grow: 0;
    }
}
</style>
```

sinisaos: Or even better without code duplication

```Vue
<template>
    <div
        id="message_popup"
        v-bind:style="[apiResponseMessage.type == 'error' ? {'background-color':'red'} : {'background-color':'green'}]"
        v-if="visible"
    >
        <p class="message">{{ message }}</p>
        <p class="close">
            <a
                class="subtle"
                href="#"
                v-on:click.prevent="visible = false"
            >
                <font-awesome-icon icon="times" />
            </a>
        </p>
    </div>
</template>
```

dantownsend: Yes, that's nice. I slightly prefer using a `:class`, just because a CSS variable can be used for the colour.

In `variables.less` there's a `@red` value - so this way, we can use that, and it keeps the colour consistent with the rest of the app.

sinisaos:
@dantownsend I pushed another commit. If that's not what you mean, fell free to change the code.

dantownsend:
@sinisaos Great, thanks!

dantownsend:
@sinisaos Perfect, thanks

Issue #99, opened 2021-10-12
Nullable timestamp fields


dantownsend:
In the UI, if there's a timestamp field and you try to save it with an empty value, you'll get a validation error.

It's because it's being sent to the API as an empty string, rather than null.

We need to modify it, so null is being sent back instead.

Issue #80, opened 2021-08-08
Nullable foreign keys


dantownsend:
There's a bug when trying to add a row with a foreign key. If the foreign key is nullable, and you try and submit an empty value, validation will fail, as Pydantic sees it as an empty string, and not None. The error returned will then be something like `a string isn't a valid type`.

The front end needs modifying to return null in these situations.

PR #100, opened 2021-10-17
fix nullable FK and timestamp fields

sinisaos:
Related to #80 and #99


dantownsend:
@sinisaos Thanks a lot for this.

PR #115, opened 2021-10-28
Hybrid select widget for picking a foreign key

### LINKED ISSUE #108: Hybrid foreign key selector
For foreign key columns, we used to use a HTML select for picking values.

We replaced it with a search box. This was necessary because the page would crash when using a select if the related table had a lot of rows. 

Screenshot 2021-10-27 at 13 27 55

In certain situations the select box is better though. For example, if the related table only has a handful of rows, it's easier to just select the one you want.

The downside of a search box is you need some prior knowledge of what you're looking for.

It might be possible to create a hybrid. When you click on the search box, and haven't typed anything, it automatically shows the first 10 possible results.

dantownsend:
Fixes #108

Screenshot 2021-10-31 at 17 38 00


dantownsend:
@sinisaos I'd appreciate your thoughts on this if you have time to check it.

You need to be running the latest version of Piccolo API for it to work.

sinisaos (approved):
Wow. This looks great. And also works great.

Screenshot from 2021-10-31 18-56-39

dantownsend:
@sinisaos Thanks a lot

dantownsend:
I'm finally going to merge this in. The problem has been with the 'null' option - it doesn't currently work with filters.

But as we're adding an operator type of 'null' to the API, we will soon be able to fix this.

Issue #121, opened 2021-10-29
Clearing filters sets the operators to 'less than'


dantownsend:
When resetting filters, the default should be 'equals' not 'less than'.

<img width="325" alt="Screenshot 2021-10-29 at 12 35 43" src="https://user-images.githubusercontent.com/350976/139427888-c550746a-a4e1-4c7c-8e2b-75f2f3bf7bb5.png">

sinisaos:
@dantownsend I see that you are really busy with the main Piccolo repo and that this bug has been assigned to you. A quick fix would be something like this.
```javascript
// in components/RowFilter.vue
async clearFilters() {
  console.log("Clearing ...")
  let form: any = this.$refs.form
  form.reset()
  let _ = [...form.elements].forEach((element) => {
    if (element.type == "hidden") {
        element.value = ""
    }
    if (element.name.includes("__operator")) {
        element.selectedIndex = 2
    }
  })
```
I can do PR if you don’t mind and if you approve this fix.

dantownsend:
@sinisaos Yeah, I assigned it to myself as I hoped to fix it before the PyData demo, but didn't have time.

There's a few temporary solutions. Something like what you suggested. Another is just to rearrange the values, so 'Equals' is at the top:

https://github.com/piccolo-orm/piccolo_admin/blob/a43f157e5fbb6b386cccb81b0d2ec6a231011e75/admin_ui/src/components/OperatorField.vue#L19-L40

I'd like to find a proper solution using Vue, but something temporary would be OK if it's not too hacky.

sinisaos:
@dantownsend You're right. It should work if we move `Equals` to 0 index in operators array, but I have to try. If it works it is an even simpler and better solution.

dantownsend:
@sinisaos Yeah, lets just try moving it to the top for now.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_admin/pull/127

PR #127, opened 2021-11-16
fixing clearing filters bug

sinisaos:
Related to #121


dantownsend:
@sinisaos Thanks for this - seems to fix the problem.

Issue #111, opened 2021-10-27
Loading indicators


dantownsend:
When the network is very slow and / or a lot of data is being pulled from the API, you can see the UI load.

Investigate whether having load spinners improves or degrades the user experience.

sinisaos:
@dantownsend Can you please close this issue because it is done.

PR #130, opened 2021-11-19
added spinner when data is loading

sinisaos:
Related to #111


dantownsend:
@sinisaos  Thanks for this - sorry it's taken me so long to leave feedback.

It looks cool.

I've added two videos here - when the network is slow it looks good. But when the network is fast, it flashes in too fast, and is a bit disorientating. If the loading spinner just covers the table instead of the whole page, this might be better. What do you think?

https://user-images.githubusercontent.com/350976/146156942-363918fc-7df7-47df-8486-87c78ff8562b.mp4

https://user-images.githubusercontent.com/350976/146157060-36e262f2-7bfa-4419-8609-2facb5ced17e.mp4

sinisaos:
@dantownsend Sorry for the late reply. You are right about loading spinners only in the table section. I will make the changes as soon as possible (probably in a day or two).

dantownsend:
@sinisaos No worries, thanks!

sinisaos:
@dantownsend Can you check what it looks like now. Feel free to add additional customizations. Thanks.

dantownsend:
@sinisaos I had a look. I think it's better now.

I tried finding a solution for when the network is very fast. There's an [API which tells you how fast the network is](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType). I was thinking the spinner could be disabled when the network is fast. This API isn't widely supported though, so isn't a great solution.

I think what helps is if the spinner / loading indicator is more subtle. In this video, I replaced the spinner with some text. At the start the network is slow, and at the end the network is fast, to show how it looks in each situation.

https://user-images.githubusercontent.com/350976/146921266-cd82f473-da99-4dee-9acc-399f64a3606f.mov

What do you think? I think something more subtle is easier on the eyes when it flashes in very quickly.

sinisaos:
@dantownsend Wow. Your version look much better. Can you please make your changes to this PR and merge to master.

dantownsend:
@sinisaos I'll merge it later today, thanks.

dantownsend:
OK, I think I finally figured out a solution to it flashing in and out too fast - I'll try putting a CSS transition on it.

Because it fades in, you won't see it if the duration is really small.

dantownsend:
This is it with the transition added. Because of the CSS transition you barely see the 'loading text' when the network is fast.

I put a transition on the table too, so it fades in. What do you think?

https://user-images.githubusercontent.com/350976/147000935-cb798d73-6628-4b49-87e5-1b4fa347143b.mov

sinisaos:
@dantownsend This is great and even better than the previous version :)

dantownsend:
@sinisaos Thanks!

PR #137, opened 2021-12-14
fix font family

dantownsend:
I only realised today that `font-family` in CSS doesn't automatically apply to `button` or `textarea` elements.

This PR adds this, and increases the font size slightly.


sinisaos (approved):
This is great.

dantownsend:
@sinisaos Thanks

Issue #140, opened 2021-12-23
Problem with left sidebar menu on small screens


sinisaos:
I didn't realise before, but there is problem with left sidebar menu on small screens after latest loading UI changes.

![sidebar](https://user-images.githubusercontent.com/30960668/147203605-c100b89e-36c4-4788-8084-15fc3ab59221.png)

dantownsend:
@sinisaos Ah, good catch - this was my fault. The sidebar just needs a z-index value adding. Will fix it now.

dantownsend:
@sinisaos I just pushed a new version to Piccolo Admin. Does this fix the issue for you?

sinisaos:
@dantownsend Yes. Issue resolved. Thanks.

dantownsend:
@sinisaos Cool, thanks a lot

PR #141, opened 2021-12-23
add z-index to sidebar

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_admin/issues/140


sinisaos:
@dantownsend I can't try it now, but I can later today and I'll let you know the result. But if this works with you, it will probably work with me.

PR #142, opened 2021-12-23
fixing z-index for modal and dropdown menu

sinisaos:
Small `z-index` fix for modal and dropdown menu.

Before:

bad

After:

good


dantownsend:
@sinisaos You're right, thanks for this.

sinisaos:
@dantownsend These are small details that can be easily missed. I hope there are no more of these little problems now.

Issue #81, opened 2021-08-08
Make sure Piccolo admin works with custom primary key columns


dantownsend:
When using custom primary key columns, for example:

```python
class MyTable(Table):
    pk = UUID(primary_key=True)
```

There are likely to be some edge cases not currently supported by Piccolo admin.

abelayalew:
Am having issue custom primary key in the admin page with page not found error

abelayalew:
okay thanks, hope it will be fixed soon, am so exited to try everything, its amazing project!!

dantownsend:
@abelayalew Thanks

cheesycod:
has this issue been fixed yet?

dantownsend:
@cheesycod Not yet unfortunately - we'll prioritise it.

sinisaos:
@dantownsend I played around with it a bit and it’s not an easy task. If I understood correctly, `PiccoloCRUD` and `FastAPIWrapper` need to be changed to accept `row_id` as a `int` (like is now) or `string` (UUID), but the whole concept of `crud` operations and `relations` between tables is based on a `Serial` type (`Integer`)?

dantownsend:
@sinisaos Yeah, you're right - the endpoints might need modifying to accept string or integers. I'll take a look to see how feasible it is.

sinisaos:
@dantownsend I managed to do a `crud` either with an `int` or with a `string`, but not both at the same time. I tried typing the `row_id` with `Any` or `Union[int, str]`, but niether worked. I didn't try FK relations.

dantownsend:
This should be fixed now if you upgrade to the latest `piccolo_api` version (0.32.3).

Apologies for the delay on this. I'll close this issue for now - if anyone encounters issues with the fix, feel free to leave more comments on this issue.

sinisaos:
@dantownsend This unfortunately doesn't work. Probably everything is fine with the newest changes in `Piccolo API`, but we will have to change the `Piccolo Admin` as well. I haven't researched too much, but the first problem is that we use the `id` column in `RowListing.vue` 
https://github.com/piccolo-orm/piccolo_admin/blob/70220411ce7bf5d785bbe02d5ffec8f4207be4f0/admin_ui/src/views/RowListing.vue#L99
This piece of code is an example that applies to every line where we use `name == 'id'` or `row.id` and if the user names the column differently, e.g. `pk` nothing will work.  

[...]

PR #143, opened 2022-01-12
made the RowListing page work with custom PK name

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_admin/issues/81

This fixes the listing page when the table has a custom PK column name.

### DIFF
--- a/admin_ui/src/components/RowSortModal.vue
+++ b/admin_ui/src/components/RowSortModal.vue
@@ -7,7 +7,7 @@
                 name="property"
                 v-model="propertyName"
             >
-                <option value="id">id</option>
+                <option :value="schema.primary_key_name">{{ schema.primary_key_name }}</option>
 
                 <option
                     :key="key"
@@ -41,7 +41,7 @@ export default {
     data() {
         return {
             ascending: "descending",
-            propertyName: "id",
+            propertyName: this.schema.primary_key_name,
         }
     },
     components: {
--- a/admin_ui/src/views/RowListing.vue
+++ b/admin_ui/src/views/RowListing.vue
@@ -86,17 +86,17 @@
                                         <th></th>
                                     </tr>
 
-                                    <tr v-bind:key="row.id" v-for="row in rows">
+                                    <tr v-bind:key="row[pkName]" v-for="row in rows">
                                         <td>
                                             <input
-                                                :value="row.id"
+                                                :value="row[pkName]"
                                                 @click="selectRow"
                                                 type="checkbox"
                                                 v-model="selectedRows"
                                             />
                                         </td>
                                         <td v-bind:key="name" v-for="name in cellNames">
-                                            <span class="link" v-if="name == 'id'">
+                                            <span class="link" v-if="name == pkName">
                                                 <router-link
                                                     :to="{
                                                         name: 'editRow',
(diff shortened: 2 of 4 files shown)


sinisaos:
@dantownsend This looks great. Just one question. Will you make changes to `EditRow.vue` and all the editing components in this PR or will you make a separate PR for it?

dantownsend:
@sinisaos Thanks. I think the outstanding issues are with the foreign key selector, but there may be others.

I'll see if I have time later to fix the others. Otherwise I'll release this partial fix.

sinisaos:
@dantownsend To solve foreign key selector we need to change this line https://github.com/piccolo-orm/piccolo_api/blob/caa8d7f21b51b6fdfdfdf527d873ded2c0316afc/piccolo_api/crud/endpoints.py#L427
in Piccolo API to this so the `api/tables/some-table/ids/` endpoint can return `primary key` as int,str or uuid and that work fine.

```python
return JSONResponse(
    {
        str(i[self.table._meta.primary_key._meta.name]): i[
            "readable"
        ]
        for i in values
    }
)
```
But that is unfortunately a minor part of the problem. As I already wrote, this is not an easy task because the whole concept of `crud` operations and `relations` between tables is based on `Serial` type, and we need to change all crud methods (that contain `row_id`) in the Piccolo API to be able to use it in Piccolo Admin.

dantownsend:
@sinisaos Yes, I see what you're saying. I think it's fixable - looking now.

dantownsend:
@sinisaos I've opened two more PRs. Combined with this one, it seems to fix the foreign key selector:

https://github.com/piccolo-orm/piccolo/pull/392
https://github.com/piccolo-orm/piccolo_api/pull/132

sinisaos:
@dantownsend Changes in the Piccolo API certainly solve the problem because I have already tried and tested it and everything work fine. Maybe it would be better to add `pk = UUID(primary_key=True)` to `Director` table because that table has a relation to `Movie` table and so you will be able to test if `FK` works. I did so. Custom `PK` in table without relations work fine, FK relation is problem.

dantownsend:
@sinisaos If you pull the latest from this branch, I've added a foreign key from Movie to Studio, so it can be tested properly.

sinisaos:
@dantownsend I already wrote this in Piccolo ORM pull request but WELL DONE AGAIN. Now with dynamic `value_type` for FK everything work great. All crud methods (without need to changes in Piccolo API) with or without FK relations, filtering etc. work great. Great. I think that will be it.

sinisaos:
@dantownsend Sorry to write like this in comment, but can you also change `RowSortModal.vue` to something like this that sorting also works with custom PK. Thanks.

```vue
<template>
    <Modal v-on:close="$emit('close')">
        <template v-if="schema">
            <p>Sort by:</p>

            <select name="property" v-model="propertyName">
                <option :value="propertyName">
                    {{ propertyName }}
                </option>

                <option
                    :key="key"
                    :value="key"
                    v-for="(property, key) in schema.properties"
                >
                    {{ key }}
                </option>
            </select>

            <select name="ordering" v-model="ascending">
                <option value="ascending">Ascending</option>
                <option value="descending">Descending</option>
            </select>

            <button v-on:click.prevent="save">Sort</button>
        </template>
    </Modal>
</template>

<script lang="ts">
import Modal from "./Modal.vue"

export default {
    props: {
        schema: Object,
        tableName: String
    },
    data() {
        return {
            ascending: "descending",
            propertyName: this.schema["primary_key_name"]
        }
    },
    components: {
        Modal
    },
    methods: {
        async save() {
            let config: i.SortByConfig = {
                property: this.propertyName,
                ascending: this.ascending == "ascending"
            }
            this.$store.commit("updateSortBy", config)

            await this.$store.dispatch("fetchRows")

            this.$emit("close")
        }
    },
    mounted() {
        let sortBy: i.SortByConfig | null = this.$store.state.sortBy
        if (sortBy) {
            this.ascending = sortBy.ascending ? "ascending" : "descending"
            this.propertyName = sortBy.property
        }
    }
}
</script>

<style>
</style>
```

dantownsend:
@sinisaos Thanks very much. I'll make the last changes you suggested and will release later tonight or in the morning.

sinisaos:
@dantownsend Great :+1: . I think everything should work. The first thing in the morning will be to check out the new version.

PR #144, opened 2022-01-16
Supporting foreign keys which reference non-primary keys

dantownsend:
Related to https://github.com/piccolo-orm/piccolo/issues/394

### DIFF
--- a/admin_ui/src/views/RowListing.vue
+++ b/admin_ui/src/views/RowListing.vue
@@ -114,7 +114,7 @@
                                             <span
                                                 class="link"
                                                 v-else-if="
-                                                    isForeignKey(name) & (row[name] !== null)
+                                                    isForeignKey(name) & (row[pkName] !== null)
                                                 "
                                             >
                                                 <router-link
@@ -122,7 +122,7 @@
                                                         name: 'editRow',
                                                         params: {
                                                             tableName: getTableName(name),
-                                                            rowID: row[name]
+                                                            rowID: row[pkName]
                                                         }
                                                     }"
                                                     >{{ row[name + "_readable"] }}</router-link
--- a/admin_ui/src/components/RowFormSelect.vue
+++ b/admin_ui/src/components/RowFormSelect.vue
@@ -35,7 +35,7 @@
                         }"
                         class="add"
                         target="_blank"
-                        v-if="!isFilter"
+                        v-if="!isFilter && getKeySelectID(property.title)"
                     >
                         <font-awesome-icon icon="edit" />
                     </router-link>
--- a/piccolo_admin/example.py
+++ b/piccolo_admin/example.py
@@ -41,7 +41,13 @@ from piccolo_api.session_auth.tables import SessionsBase
 from pydantic import BaseModel, validator
 
 from piccolo_admin.endpoints import FormConfig, TableConfig, create_admin
-from piccolo_admin.example_data import DIRECTORS, MOVIE_WORDS, MOVIES, STUDIOS
+from piccolo_admin.example_data import (
+    DIRECTORS,
+    MOVIE_WORDS,
+    MOVIES,
+    REVIEWS,
+    STUDIOS,
+)
 
 
 class Sessions(SessionsBase):
@@ -75,7 +81,7 @@ class Director(Table, help_text="The main director for a movie."):
 
 class Studio(Table, help_text="A movie studio."):
     pk = UUID(primary_key=True)
-    name = Varchar()
+    name = Varchar(unique=True)
     facilities = JSON()
 
 
@@ -90,7 +96,7 @@ class Movie(Table):
         romance = 7
         musical = 8
 
-    name = Varchar(length=300)
+    name = Varchar(length=300, unique=True)
     rating = Real(help_text="The rating on IMDB.")
     duration = Interval()
     director = ForeignKey(references=Director)
@@ -103,6 +109,21 @@ class Movie(Table):
     barcode = BigInt(default=0)
     genre = SmallInt(choices=Genre, null=True)
     studio = ForeignKey(Studio)
+
+    @classmethod
+    def get_readable(cls):
+        return Readable(template="%s", columns=[cls.name])
+
+
+class Review(Table):
+    reviewer = Varchar()
+    content = Text()
+    rating = Integer()
+    movie = ForeignKey(Movie, target_column=Movie.name)
+
+    @classmethod
+    def get_readable(cls):
+        return Readable(template="%s %s", columns=[cls.reviewer, cls.rating])
 
 
 class BusinessEmailModel(BaseModel):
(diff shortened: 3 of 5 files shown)


sinisaos:
@dantownsend We need to change this line 
https://github.com/piccolo-orm/piccolo_admin/blob/58b68387e725654791b9c3ffa1f4902905f594c5/admin_ui/src/views/RowListing.vue#L125
to this `rowID: row[pkName]`

and this line
https://github.com/piccolo-orm/piccolo_admin/blob/58b68387e725654791b9c3ffa1f4902905f594c5/admin_ui/src/components/RowFormSelect.vue#L33
to this `rowID: getKeySelectID(schema.primary_key_name)`

With these two changes, everything is fine, and all that remains is to fix filter FK search. Of course, this needs to be double checked.

dantownsend:
@sinisaos Great, thanks. I'll try those changes.

sinisaos:
@dantownsend Forget about these changes because they are not good and they not work properly. Sorry.

sinisaos:
@dantownsend Sorry, but these changes don't work either. I can make the links work, but I need to change this https://github.com/piccolo-orm/piccolo_api/blob/f0db58c9961c8a182b35d71769d99aea5abbf190/piccolo_api/crud/endpoints.py#L872
to something like this
```python
try:
    try:
        row_id = self.table._meta.primary_key.value_type(row_id)
    except ValueError:
        for i in self.table._meta._foreign_key_references:
            reference_target_pk = (
                await self.table.select(self.table._meta.primary_key)
                .where(i._meta.params["target_column"] == row_id)
                .first()
                .run()
            )
            row_id = reference_target_pk[self.table._meta.primary_key]
  except ValueError:
    return Response("The ID is invalid", status_code=400)
```
in the Piccolo API. Changes in Piccolo Admin are not enough.

admin_ui/src/views/RowListing.vue
```diff
@@ -114,15 +114,15 @@
...
+                                                    isForeignKey(name) & (row[pkName] !== null)
                                                 "
                                             >
                                                 <router-link
                                                     :to="{
                                                         name: 'editRow',
                                                         params: {
                                                             tableName: getTableName(name),
-                                                            rowID: row[name]
+                                                            rowID: row[pkName]
```
sinisaos, line 125: I also thought so before, but the `rowID: row[pkName]` is the primary key of the `Review` table, not `Review.movie.id`.
I wrote in the comment the changes I think we need to make in Piccolo API to get correct results.

dantownsend:
@sinisaos Yeah, it's a tricky one. I had a look at it earlier, but still haven't fully wrapped my head around it yet.

sinisaos:
@dantownsend I agree that it is tricky and I hope there is a simpler way, but for now I do not see it. The changes in my first attempt and in your last commit are not good because they do not work both for primary key and non-primary foreign keys. Eventually I made the changes from the previous comment (with this change we get the primary key of table where the `target_column` is located) to `detail` endpoint in Piccolo API crud and it works well for both. Today I found a potential solution for the filter search sidebar by making changes to `ids` endpoint and also for Piccolo Admin. If you interested, I can write these changes here in the comment or I can made PR in both repos so you can try this. But first we need to fix tests in Piccolo API after new Piccolo ORM features. We need to add `target_column` to schema response in `test_crud_enpoints.py` and `test_fastapi.endpoints.py` and fix session auth test because `ValueError: Do not pass hashed password.` is raised in 4 tests due to new BaseUser `from_fixture` changes.

dantownsend:
@sinisaos Damn - good catch. I didn't realise the BaseUser changes broke Piccolo Admin. Will look into fixing it ASAP.

sinisaos:
@dantownsend You need to change session auth Piccolo API tests.

dantownsend:
@sinisaos Yeah, you're right. Seems like there's an error when accessing `request.user` with session middleware.

dantownsend:
@sinisaos I need to partially revert some of the changes to `BaseUser`.

dantownsend:
@sinisaos I've pushed a new release of Piccolo which solves the underlying problem.

If you have time, are you able to do a fix for Piccolo API. Looks like the errors are because 'target_column' is now in the schema, and the tests haven't been updated accordingly.

dantownsend:
> Today I found a potential solution for the filter search sidebar by making changes to `ids` endpoint and also for Piccolo Admin

I'm thinking along the same lines. I was thinking we could pass a `target_column` GET param to the ids endpoint, and then it would modify its response accordingly. For example:

```
GET /ids/?target_column=name
```

Would return something like:

```
{
'Star Wars': 'Star Wars'
}
```

sinisaos:
> If you have time, are you able to do a fix for Piccolo API? Looks like the errors are because 'target_column' is now in the schema, and the tests haven't been updated accordingly.

I can’t right now but I will do it as soon as I get home during the day. You are right, to fix test errors you just need to add target column to response schema.

sinisaos:
> @sinisaos I'm thinking along the same lines. I was thinking we could pass a `target_column` GET param to the ids endpoint, and then it would modify its response accordingly. For example:
> 
> ```
> GET /ids/?target_column=name
> ```
> 
> Would return something like:
> 
> ```
> {
> 'Star Wars': 'Star Wars'
> }
> ```

That's a good idea, I'll have to try that. My solution is that if the primary key is not `int` or `uuid` then I add boolean to the `ids` dict value list, something like this
```
{
     1: ['Star Wars', True],
}
```
and then in Piccolo Admin make for example
`this.hiddenSelectedValue = event.readable[0] ? event.readable[1] == true: event.id`
I don't know exactly how it goes because I don't see the code at the moment, but with this changes everything work fine both primary or non-primary keys. I can do PR in both repoes or even simpler I can just write all the changes here (there are not many changes) in the comment so you can try it and say your opinion.

sinisaos:
@dantownsend I made [PR](https://github.com/piccolo-orm/piccolo_api/pull/134) in Piccolo API and with this changes you have to change 2 files in Piccolo Admin 
```javascript
// in KeySearch.vue
methods: {
    handleUpdate(event) {
        this.selectedValue = event.readable[0]
        this.hiddenSelectedValue =
            event.readable[1] == false ? event.id : event.readable[0]
        this.showModal = false
    }
},
```
and
```javascript
// KeySearchModal.vue
<li :key="id[0]" v-for="id in ids">
    <a href="#" @click.prevent="selectResult(...id)">
        {{ id[1][0] }}
    </a>
</li>
```
What do you think about all this?

admin_ui/src/views/RowListing.vue
dantownsend: You're right - it took a while for me to realise this. My solution doesn't work.

sinisaos:
@dantownsend You should close this PR. Fix is [​​here](https://github.com/piccolo-orm/piccolo_api/pull/134).

sinisaos:
@dantownsend You should close this PR. Fix is [here](https://github.com/piccolo-orm/piccolo_api/pull/134).

Issue #152, opened 2022-03-14
Wysiwyg editor


sinisaos:
@dantownsend are you perhaps interested in adding wysiwyg editor to textarea fields?

https://user-images.githubusercontent.com/30960668/158114813-3fa9b405-e87e-438e-84a1-5a2395cc145e.mov

dantownsend:
@sinisaos It's a nice idea. The widget fits in well.

I think something like this should be configurable, rather than always on. In ``TableConfig`` there might be a rich text option or something.

```python
TableConfig(Movie, rich_text_columns=[Movie.content])
```

What widget is it?

sinisaos:
@dantownsend It is a [Vue2Editor](https://www.vue2editor.com/) based on [Quill](https://quilljs.com/). This feature is only for textarea fields and I think the boolean value is enough

```python
TableConfig(Movie, rich_text_widget=True) # default value is false and rich_text_widget is optional
```
If you want I can do PR?

dantownsend:
It's more convenient having a boolean, but I can imagine situations where you only want it for a subset of fields.

Imagine you had a blog post table. You might have a column for the description, which is plain text, and a column for the post contents, which is rich text.

What about:

```python
TableConfig(Movie, rich_text_columns=[Movie.content])

# OR, for all:
TableConfig(Movie, rich_text_columns=['__all__'])
```

sinisaos:
@dantownsend Makes sense. I will try to do that.

dantownsend:
@sinisaos Thanks. I think my only concern is it could lock us into Vue 2. I wonder if there are any widgets which work on Vue 2 and Vue 3?

sinisaos:
@dantownsend I don't know when you plan to switch to Vue 3, but I don't think there should be a problem with any rich text editor (I've tried it with CKEditor and it also works) because when we save form data, we actually use data from the hidden textarea field, and not from the widget. Something like this

```javascript
// InputField.vue
<div v-else-if="format == 'text-area' && isFilter == false">
    <vue-editor
        v-model="localValue"
        v-bind:name="getFieldName(title)"
        :editor-toolbar="customToolbar"
    />
    <textarea   //  we use data from this hidden textarea field
        id="editor"
        v-model.lazy="localValue"
        v-bind:name="getFieldName(title)"
    ></textarea>
</div>

<style scoped lang="less">
textarea#editor {
    display: none;
}
</style>
```

dantownsend:
[...]

PR #153, opened 2022-03-15
adding wysiwyg editor

sinisaos:
Related to #152


dantownsend:
Cool, thanks for this.

sinisaos:
I always forget about dark mode :). I didn't find a dark mode for Quill, so I will have to find svg css classes and change them for dark mode. 

> Other than that my only concern is it locks us to Vue 2. When we move to Vue 3 we will have to find a different library.

Because we use the hidden textarea to submit content, everything will be fine and the transition should not be difficult. We only need to find widget for Vue3 that suit us. The are several option like several options like [CKEditor](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/vuejs-v3.html) or [vue-quill](https://github.com/vueup/vue-quill).

sinisaos:
Is this okay?

dark

dantownsend:
@sinisaos Yes, that looks great

sinisaos:
@dantownsend Can you add docs for `rich_text_columns` with screenshots? I don't have a Safari and it would be nice if the screenshots is from the same browser.

dantownsend:
Yeah, I'll try and add the docs and screenshots today.

dantownsend:
I had a look at the new bundle size with the WYSIWYG library included - chunk-vendors.js went up around 300 kB (ignore the blue highlighted lines):

Screenshot 2022-04-03 at 20 17 41
Screenshot 2022-04-03 at 20 15 51

It might be worth doing some testing on Chrome with network throttling enabled. What do you think?

There are some [smaller editors](https://github.com/jaredreich/pell#comparisons), but they don't seem as nice as Quill.

sinisaos:
@dantownsend Chromium DevTools results:
```
Without wysiwyg editor (hard refresh - no cache) size is 300 kb:
Mid-tier mobile 3.12 s
Low-end mobile 11.22 s

With wysiwyg editor (hard refresh - no cache) size is 600 kb:
Mid-tier mobile 4.69 s
Low-end mobile 16.79 s
```
Feel free to implement another editor (Pell have also [Vue implementation](https://github.com/jaredreich/pell/blob/v2/examples/vue.md)).I basically chose vue2-editor because it is easy to implement, and Quill is among the most popular and most frequently used editors. Rich Text Editor is just an optional feature I've seen with other admin interfaces.

dantownsend:
@sinisaos Thanks for checking that. 

Let's just release it. I think it's a good feature, and we can always swap in a different editor in the future.

Issue #163, opened 2022-06-02
add filtering to url so that back button works as expected


trondhindenes:
If I'm in a filtered list, and then click an item for details and then click back, I'm taken to the unfiltered view. It would be good if filters were expressed using url query params so that back navigation would take me back to the filtered view.

sinisaos:
@trondhindenes There is already a [issue](https://github.com/piccolo-orm/piccolo_admin/issues/42) with that problem, but progress has somehow stalled.

dantownsend:
Yeah, I agree this needs adding. I've increased the priority of the issue.

Issue #42, opened 2021-03-17
Synchronise the filters with the GET parameters


dantownsend:
There's a little known feature in Piccolo admin where you can filter the row listing page using GET params. For example:

```
/#/movie?won_oscar=true
```
However, these values aren't synchronised with the filter sidebar:

 * When the page loads, the filter sidebar doesn't display the values in the GET params
 * When the filter sidebar values change, the GET params don't update accordingly

Having this work properly would be a nice feature. It will allow a user to do the following:

 * Load the listing page, and apply some filters
 * Click on one of the results, and edit it
 * Press the back button, and the same filters will still be applied

PR #165, opened 2022-06-02
synchronise the filters with the GET parameters

sinisaos:
Related to #42 and #163


sinisaos:
@dantownsend Did you have time to try this?

dantownsend:
@sinisaos Not yet, will have a look today.

dantownsend:
This is cool.

I came across an issue though - even though the values in the URL are applied as filters, they don't show in the filter sidebar for some reason.

https://user-images.githubusercontent.com/350976/172550468-d74f7d72-06a0-483e-9005-d9d3f19a1035.mov

I'm not sure why - will have to investigate more.

sinisaos:
@dantownsend You're right. Sorry, I didn't watch it at all. I focused on being able to get back to the filtered view after pressing `Back` button.

sinisaos:
@dantownsend Can you try this with last commit? It's keeping the form data after changing the route. Work with input type text and number, doesn't work on select.

dantownsend:
Sorry I haven't had a change to look into this yet - will get it in the next release.

sinisaos:
@dantownsend Another try with [history.state](https://developer.mozilla.org/en-US/docs/Web/API/History/state). The main problem is that the duration and keySearch widget are not in sync with `history.state`, so I changed that. The code is not the prettiest, but it works. Fell free to change that.


https://user-images.githubusercontent.com/30960668/180741650-a18b8785-dda9-48e1-a668-c8a8276eb299.mp4

dantownsend:
@sinisaos Cool, thanks - it looks like it works well in the video.

sinisaos:
@dantownsend It works, but the code is not the nicest.

dantownsend:
Still valid - just need to test it.

sinisaos:
Unfortunately, this no longer works. Need to find another way to synchronize the query parameters and the filter form

Issue #158, opened 2022-04-18
Add bulk update option


dantownsend:
If would be great if we had a bulk update option, where you select some rows, then can set a value on all of those rows. For example, with blog posts, setting them all from 'draft=True' to 'draft=False'.

sinisaos:
@dantownsend If you don't mind we can use Javascript `prompt()` to easy achieve this. Very similar to bulk delete. Something like this:
```javascript
// in RowListing.vue
methods: { 
    ...
    async updateRows() {
        let data = {}
        let column = prompt(
            "Enter the name of the column you want to update"
        )
        data[column] = prompt("Enter the value you want to update")
        for (let i = 0; i < this.selectedRows.length; i++) {
            await this.$store.dispatch("updateRow", {
                tableName: this.tableName,
                rowID: this.selectedRows[i],
                data: data
            })
        }
        await this.fetchRows()
    },
    ...
}
```
If you think this idea is acceptable, I can do PR to see all the code.

sinisaos:
@dantownsend What do you think about this?

dantownsend:
Implemented by https://github.com/piccolo-orm/piccolo_admin/pull/166

PR #166, opened 2022-06-10
bulk update

sinisaos:
Related to #158


sinisaos:
@dantownsend Result is:


https://user-images.githubusercontent.com/30960668/174348541-e5b4811f-3eae-46f9-8337-375dd8ff795e.mp4

dantownsend:
This is interesting.

I think it works in most situations, but might confuse a non-technical user in these cases:

 * Let's say you're updating the gender, which is a choices field - even though it shows 'male' in the UI, to update it to female you actually send back 'f'.
 * If the column name is 'Oscar Nominations', you need to make sure you call it 'oscar_nominations'

Ideally, we would show a modal like this where the user can select the field and value:

piccolo_admin_bulk_update

It would look very similar to the 'order' modal.

sinisaos:
@dantownsend You're right. A modal would be a much better solution. I will try to change that. 

P.S. when we have `bulk_update` from Piccolo API PR we will no longer need to use loop through `patch_single` endpoint but we will have proper bulk update on the backend side and then we will also change `updateRows` function in Piccolo Admin.

dantownsend:
That looks really cool.

Screenshot 2022-07-06 at 22 08 31

I don't think we should list the primary key value in the list, as they shouldn't be allowed to change that, especially in bulk.

It looks like when the modal loads, it shows 'value to change' with no input underneath. Perhaps we can hide 'value to change' until they've selected a column?

admin_ui/src/components/BulkUpdateModal.vue
```diff
@@ -0,0 +1,129 @@
...
+                </option>
+                <option
+                    :key="key"
+                    :value="key"
+                    v-for="(property, key) in schema.properties"
+                >
+                    {{ key }}
+                </option>
+            </select>
+            <p v-show="propertyName != schema.primary_key_name">
```
dantownsend, line 18: I'd change this to a `v-if` - `v-show` still adds it to the DOM, but hidden.

docs/source/actions/index.rst
```diff
@@ -0,0 +1,30 @@
...
+Ability to update multiple rows in one action. Select which rows you
+want to update by checking the checkboxes. By pressing the ``update``
+button you will be asked to pass the column name and the value which want to 
+update in bulk.
+
+
+Export to csv
+-------------
+
+Export data to a csv document by pressing the ``export csv`` button.
```
dantownsend, line 27: `csv` -> `CSV`

docs/source/actions/index.rst
```diff
@@ -0,0 +1,30 @@
...
+Ability to delete multiple rows in one action. Select which rows you
+want to delete by checking the checkboxes and pressing the ``delete`` button.
+
+
+Bulk update
+-----------
+
+Ability to update multiple rows in one action. Select which rows you
+want to update by checking the checkboxes. By pressing the ``update``
+button you will be asked to pass the column name and the value which want to
```
dantownsend, line 20: `which want` -> `which you want`

admin_ui/src/components/BulkUpdateModal.vue
```diff
@@ -0,0 +1,129 @@
...
+        schema: Object,
+        tableName: String,
+        selectedRows: {
+            type: Array,
+            defult: () => []
+        },
+        isFilter: {
+            type: Boolean,
+            default: false
+        }
```
dantownsend, line 57: We can probably drop this prop - I don't think it's needed.

On `InputField`, just set `isFilter` to `false`. 

```vue
<InputField
    v-if="key == propertyName"
    v-bind:format="property.format"
    v-bind:isFilter="false"
    v-bind:key="property.title"
    v-bind:title="property.title"
    v-bind:type="property.type || property.anyOf[0].type"
    v-bind:choices="property.extra.choices"
/>
```

admin_ui/src/components/BulkUpdateModal.vue
```diff
@@ -0,0 +1,129 @@
...
+    },
+    components: {
+        Modal,
+        InputField
+    },
+    data() {
+        return {
+            propertyName: this.schema.primary_key_name
+        }
```
dantownsend, line 66: I"m not sure about this - setting it to the primary key name is a bit confusing. Can it just default to null instead?

admin_ui/src/components/BulkUpdateModal.vue
sinisaos: You're right. I will change that.

admin_ui/src/components/BulkUpdateModal.vue
sinisaos: We can't change that because we need that value to be able to hide the `Value to change`.
```Vue
<p v-show="propertyName != schema.primary_key_name">
    Value to change
</p>
```

dantownsend: Will this work:

```vue
<p v-show="propertyName">
    Value to change
</p>
```

sinisaos: Yes. That works if I set `propertyName: null` but I need to remove `Please select column` option.

dantownsend:
@sinisaos I added a few more features to the update modal - it now supports foreign key fields. What do you think?

sinisaos:
@dantownsend This is great, especially the part with FK and nullable columns, really cool. Feel free to merge this. 

When you merge the bulk PR from the Piccolo API, I will make changes in Piccolo Admin to update to those new features.

PR #168, opened 2022-06-18
change password and new user code

sinisaos:
Ability to change password and create a new user.


admin_ui/src/components/ChangePassword.vue
```diff
@@ -0,0 +1,101 @@
...
+                console.log("Request failed")
+                console.log(error.response)
+                this.$store.commit("updateApiResponseMessage", {
+                    contents:
+                        "The form has errors. Changing password canceled.",
+                    type: "error"
+                })
+                setTimeout(() => {
+                    this.$router.push({ name: "home" })
+                }, 5)
```
dantownsend, line 78: If it fails, we should probably just keep them on the same page in case they just made a mistake, rather than sending them to the homepage.

admin_ui/src/components/ChangePassword.vue
```diff
@@ -0,0 +1,101 @@
...
+    components: {
+        BackButton
+    },
+    methods: {
+        async changePassword() {
+            console.log("Changing password")
+            const payload = {
+                old_password: this.oldPassword,
+                new_password: this.newPassword,
+                confirm_password: this.confirmPassword
```
dantownsend, line 56: I changed this parameter to `confirm_new_password` in the latest version to be more explicit.

admin_ui/src/components/ChangePassword.vue
```diff
@@ -0,0 +1,101 @@
...
+            }
+            try {
+                await axios.post(`./auth/change-password/`, payload)
+                this.$store.commit("updateApiResponseMessage", {
+                    contents: `Changed password successfully. You will be redirected 
+                        to the login page to log in with a new credentials.`,
```
dantownsend, line 62: `with a new credentials` -> `with your new credentials`

admin_ui/src/components/CreateNewUser.vue
```diff
@@ -0,0 +1,101 @@
...
+                setTimeout(() => {
+                    this.$router.push({ name: "home" })
+                }, 5)
+                return
+            }
+        }
+    }
+}
+</script>
+
```
dantownsend, line 85: I guess this is a work in progress, as it uses the change password endpoint at the moment.

We need to be careful with letting the user register other users. We should only allow it if the user has `superuser=True`.

sinisaos: Wrong copy/paste. I fixed that.

docs/source/asgi/examples/starlette/app.py
```diff
@@ -7,7 +7,13 @@
 
 # The `allowed_hosts` argument is required when running under HTTPS. It's
 # used for additional CSRF defence.
-admin = create_admin([Director, Movie], allowed_hosts=["my_site.com"])
+admin = create_admin(
+    [Director, Movie],
+    allowed_hosts=["my_site.com"],
+    # Required for change password
+    # and create new user endpoints:
+    prefix_path="/admin/",
```
dantownsend, line 15: I see the issue here. We need to redirect the user to the login page, but we don't know what URL the admin is mounted at. It means we have to do soething like:

```python
app = FastAPI()

app.mount('/admin/', create_admin([Director, Movie], prefix_path="/admin/")
``` 

Repeating the path is a bit unfortunate. I wonder if there's another way around it?

As we're calling the endpoints via REST, can we just ignore the redirect coming from the server, and just redirect from the front end instead?

piccolo_admin/endpoints.py
```diff
@@ -388,6 +392,20 @@ def __init__(
             ),
         )
 
+        auth_app.mount(
+            path="/change-password/",
+            app=AuthenticationMiddleware(
+                change_password(login_url=f"{self.prefix_path}auth/login/"),
+                SessionsAuthBackend(),
+            ),
+        )
```
dantownsend, line 401: I would consider mounting this on the `api_app` as it already has authentication enabled.

piccolo_admin/endpoints.py
```diff
@@ -388,6 +392,20 @@ def __init__(
...
+
+        auth_app.add_route(
+            path="/register/",
+            route=register(redirect_to=f"{self.prefix_path}auth/login/"),
+            methods=["POST"],
+        )
```
dantownsend, line 407: We need to be really careful about this - we don't want anybody to be able to register. I think this needs to be mounted to the `api_app`, so it requires authentication. Then only existing admin users can register new users.

docs/source/asgi/examples/starlette/app.py
sinisaos: Yes. This is a problem and I don't know how to properly solve it. If you have any ideas feel free to apply them.

piccolo_admin/endpoints.py
sinisaos: Now I have to go, but I will try to solve it during the day.

piccolo_admin/example.py
```diff
@@ -176,8 +168,8 @@ async def booking_endpoint(request, data):
     Director,
     Movie,
     Studio,
-    User,
-    Sessions,
+    BaseUser,
+    SessionsBase,
```
dantownsend, line 172: I think the reason this wasn't working is this:

Screenshot 2022-06-18 at 08 56 26

The `session_table` and `auth_table` arguments weren't being passed to `SessionsAuthBackend`. I think if that is changed, we can use the `User` and `Sessions` table as before.

piccolo_admin/endpoints.py
dantownsend: No worries - catch you later.

piccolo_admin/example.py
sinisaos: Unfortunately, this doesn't work because if the `Users` and `Sessions` table is in `example.py` I can't import them into `endpoints.py` (`ImportError` raised - circular import)

docs/source/asgi/examples/starlette/app.py
sinisaos: Somehow I managed to solve it with this:
```python
route=change_password(login_url="./../../auth/login/"),
```
It's not the nicest, but it works and we don't have to use `prefix_path` anymore in the `create_admin` function.

dantownsend: That's better. Does the `login_url` matter at all though? We're not using that HTML form - just via the API.

sinisaos: This is important because if we leave the default `login_url` of the password change endpoint, we will be redirected to `https://some-url/login`, but we change the admin user password and only the admin (`admin_only=True` in `auth_middleware`) can do that. For example
```python
app = FastAPI()

app.mount('/admin/', create_admin([Director, Movie]) 
```
need `https://some-url/admin/auth/login` not `https://some-url/login` to properly work.
If we change mount path to e.g. `piccolo-admin` we need `https://some-url/piccolo-admin/auth/login` not `https://some-url/login`. I hope that makes sense.

admin_ui/src/components/CreateNewUser.vue
sinisaos: Yeah. That makes sense. Thanks. I just have to change this.
```python
api_app.mount(
    path="/register/",
    app=AuthenticationMiddleware(
        register(redirect_to="./../../auth/login/"),
        SessionsAuthBackend(superuser_only=True),
    ),
)
```
I'll do it later today.

dantownsend: Cool - thanks

dantownsend: That's better. I think we should grey out and disable the 'create user' button in the UI if they're not super users. Otherwise they may try and create a user, and it will fail.

Screenshot 2022-06-21 at 16 51 36

piccolo_admin/endpoints.py
```diff
@@ -364,6 +366,28 @@ def __init__(
...
+                register(
+                    redirect_to="./../../auth/login/", auth_table=auth_table
+                ),
+                SessionsAuthBackend(
+                    auth_table=auth_table,
+                    session_table=session_table,
+                    superuser_only=True,
+                ),
+            ),
+        )
```
dantownsend, line 389: We need to be careful with these, because `create_admin` has a `read_only` parameter. We need a way of making these read only too. We can either modify the `register` and `change_password` endpoints to accept a `read_only` parameter, or return a generic endpoint which returns a 405 status code.

dantownsend: I've added a PR to adding read only modes to these endpoints:

https://github.com/piccolo-orm/piccolo_api/issues/156

sinisaos: Great. Thanks

admin_ui/src/components/CreateNewUser.vue
sinisaos: I've added a commit for this which you can check out.

admin_ui/src/views/ChangePassword.vue
```diff
@@ -0,0 +1,99 @@
...
+                console.log("Request failed")
+                console.log(error.response)
+                this.$store.commit("updateApiResponseMessage", {
+                    contents:
+                        "The form has errors. Changing password canceled.",
+                    type: "error"
+                })
+                setTimeout(() => {
+                    this.$router.push({ name: "changePassword" })
+                }, 5)
```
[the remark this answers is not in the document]
sinisaos: This prevents redirection to the login page (redirection to the login page occurs even if there are errors in the form when we submit the form) and we remain on the same page, but without persisted form data. If you have any ideas, feel free to change this.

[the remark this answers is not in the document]
dantownsend: Good point - axios might be intercepting any errors and redirecting to the login page. I'll have a look.

dantownsend:
One challenge we have at the moment is when creating a user, you can't set whether they're active / admin / superuser.

Screenshot 2022-06-23 at 21 32 41

In `example.py` you've registered `BaseUser` with the admin, so someone could go to the list view and set these values. However, the problem is anyone could modify it (not just superusers).

Screenshot 2022-06-23 at 21 33 26

I added something to `PiccoloCRUD` ages ago called validators:

 * https://piccolo-api.readthedocs.io/en/latest/crud/piccolo_crud.html#piccolo_api.crud.endpoints.Validators

So we can stop certain users from performing certain actions on `PiccoloCRUD`. This might be one work around - we automatically add some extra validators for `BaseUser`.

Another option is to modify the register endpoint so superusers can specify the admin / active / superuser values. There's kind of a mismatch at the moment - the register endpoint was intended for people self registering, but in the admin we're registering users on their behalf.

Sorry for the long comment - do you see what I'm saying?

sinisaos:
Good point. We can hide the `BaseUser` table from `TableNav.vue` if the user is not a superuser similar to disabling user creation and this is not a problem, but I find one big problem. When we register `BaseUser` in `create_admin`, we can perform crud operations, however if for example we want to add some privilege to the user or change some other column, with row editing the password is returned as empty string (with `Varchar` column was everything ok, but `Secret` column is not displayed in admin and when editing a column in admin with `Secret` column or `secret=True` admin UI return empty string after form submit and that is a problem) and that user can no longer log in because his password is empty string.

password

dantownsend:
@sinisaos Yes, that's also a problem. One solution is to not send back a value for a secret field when doing a PATCH request. Alternatively, there might just have to be some custom UI for listing / editing users, which incorporates a proper change password form.

sinisaos:
@dantownsend We could easily fix this by setting `BaseUser` password column as `Varchar` and everything would be fine, but that's not the solution because we have to enable proper editing of any table that has columns with `secret=True` because it doesn't work properly at the moment either. 
I’m going to try to show the secret in an edit form somehow so maybe I'll come up with something.

sinisaos:
@dantownsend I changed my approach to this. I will completely remove the `create_user` endpoint from Piccolo Admin (because you are right that the `register` endpoint is good for the end user, but not for administrators) and with some changes in Piccolo API we will be able to create and edit a users in Piccolo Admin in its existing form via PiccoloCRUD, but with the ability to properly create hashed passwords. Only the `superuser` will be able to do this, and for all others `User` table will not be displayed at all in the admin UI. I have to write some tests, but I will do it tomorrow. I'll do PR for the Piccolo API and Piccolo Admin changes so you can see what it's all about and give your opinion

dantownsend:
@sinisaos Thanks. I agree - I don't think we can use the register endpoint. Either we use PiccoloCRUD somehow, or build a new endpoint for this specific purpose.

sinisaos:
@dantownsend When you have time, create a patch for the local Piccolo API with the code from this [PR](https://github.com/piccolo-orm/piccolo_api/pull/158), and then you can try new changes in Piccolo Admin. Hopefully this is better than the previous `create_user` endpoint.

piccolo_admin/endpoints.py
```diff
@@ -516,11 +527,24 @@ def get_meta(self) -> MetaResponseModel:
...
+            .first()
+            .run_sync()
+        )
+        if user["superuser"] is True:
+            return [i.table_class._meta.tablename for i in self.table_configs]
+        return [
+            i.table_class._meta.tablename
+            for i in self.table_configs
+            if i.table_class != self.auth_table
+        ]
```
dantownsend, line 547: I see what you're doing here - we're hiding the user table from the UI.

It will still exist in the API though, so someone could potentially exploit it. Also, there will be times when a user needs to see the user table, but just not create / edit users. I'll see if we can fix this using validators, and if so will push a commit.

dantownsend:
@sinisaos I've added validators to show you what I'm talking about.

At the moment, the validators are pretty aggressive - they prevent non superusers from adding/editing/deleting.

sinisaos:
@dantownsend This is great and work with latest Piccolo API PR. 
create
and
edit
If you have time, see the Piccolo API [PR](https://github.com/piccolo-orm/piccolo_api/pull/158). I'm interested in your opinion. Exceptions still need to be added if that code suits you

Issue #169, opened 2022-06-24
Feature Request: Make custom form message timeout configurable


ethagnawl:
I'm using a custom form to allow admins to run a reporting routine and the output _can_ be pretty lengthy. It looks like the [current timeout](https://github.com/piccolo-orm/piccolo_admin/blob/569245afb0c2b6f0ae9bdc54df3f2f6032f7df0a/admin_ui/src/components/MessagePopup.vue#L51-L61) is 3 seconds and that's not enough time for the user to comprehensively review the output.

Would you be open to adding an optional property to the CustomForm class that would allow users to configure this -- and potentially other -- UI options?

dantownsend:
I'm not against adding new config options to `FormConfig`.

I gave it a try with a really slow form endpoint to see what happens. The form gets processed OK. It would probably be better if it said 'saving' or something, and disabled the submit button so you can't accidentally submit the form multiple times.

https://user-images.githubusercontent.com/350976/175757226-28218f6b-4eb5-4648-967e-1a1749153c73.mov

I think I understand your issue though - if the endpoint returns a lengthy message that the user needs to read, they might not be able to read if because the message only shows for 3 seconds.

Maybe we could change the design, so rather than just showing the message in a popup, we redirect them to a success page with the message on, which has no timeout.

<img width="382" alt="Screenshot 2022-06-25 at 05 09 48" src="https://user-images.githubusercontent.com/350976/175757673-865991e8-f58e-4ad6-8ef6-fc9b13589aa7.png">

That way the form could even return a bunch of JSON or something that the user could copy.

ethagnawl:
Thanks, yet again, for the prompt response, @dantownsend! 

> if the endpoint returns a lengthy message that the user needs to read, they might not be able to read if because the message only shows for 3 seconds.

That's exactly it. I also think your solve is an even better one that what I'd outlined. Though, I could still make the case for making the CustomForm UI more configurable (the timeout mentioned above, styles, input validation, etc.).

Let me know if you want help with a PR for the "success" page approach. (How would you envision errors being treated? as they currently are or as a message on this new page?)

dantownsend:
@ethagnawl  If you want to give it a go that would be great. I'm happy to help, or finish if off if you hit any problems.

I think for errors, it's probably OK as it is for now.

dantownsend:
@ethagnawl It looks good.

[...]

PR #171, opened 2022-06-28
show results on custom form page

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_admin/issues/169

If someone wants to returns a lot of data from a custom form, the current UI doesn't accommodate this very well (the data is just shown in a little popup at the bottom of the screen).

For example, you might have a form which returns a list of users.

The results are now shown on the form page, which makes it easier for the user to see the results, and copy them if they want.


dantownsend:
@sinisaos What do you think of this? Can you review it? I couldn't add you as a reviewer for some reason.

admin_ui/src/components/FormAdd.vue
```diff
@@ -6,12 +6,30 @@
...
+            <p>{{ successMessage }}</p>
+            <ul>
+                <li>
+                    <a href="#" @click.prevent="resetForm">Use again</a>
+                </li>
+                <li>
+                    <router-link to="/">Back to home page</router-link>
+                </li>
+            </ul>
+        </div>
```
sinisaos, line 19: I would show only a success message  to emphasize this a little (not form name and description). Something like this
```Vue
<template v-if="formConfig && !successMessage">
    <h1>{{ formConfig.name }}</h1>

    <p v-if="formConfig.description">{{ formConfig.description }}</p>
</template>

<div v-show="successMessage">
    <p id="success">{{ successMessage }}</p> <-- let it be bold and @green to be highlighted or something
    <ul>
        <li>
            <a href="#" @click.prevent="resetForm">Use again</a>
        </li>
        <li>
            <router-link to="/">Back to home page</router-link>
        </li>
    </ul>
</div>
```
We also need to add something like this to change the style of the success message
```less
<style scoped lang="less">
@import "../vars.less";

h1 {
    text-transform: capitalize;
}
#success {
    color: @green;
    font-weight: bold;
    font-size: 1.2rem;
}
</style>
```
success_message

admin_ui/src/components/FormAdd.vue
```diff
@@ -26,24 +44,28 @@ const BASE_URL = process.env.VUE_APP_BASE_URI
...
         async formSlug() {
             await this.fetchFormConfig()
-        },
+        }
     },
     methods: {
+        resetForm() {
+            this.$refs.form.reset()
+            this.successMessage = null
+        },
```
sinisaos, line 68: We need to set `this.errors = ""` because if we first have an error in the form, then we successfully submit the form and then press `Use again` the previous error will be displayed in a new form. Something like this
```vue
resetForm() {
    this.$refs.form.reset()
    this.successMessage = null
    this.errors = ""
},
```

sinisaos (approved):
@dantownsend This looks great. I have added two comments on improving the display of the `successMessage`.

admin_ui/src/components/FormAdd.vue
dantownsend: Good point - will change it.

admin_ui/src/components/FormAdd.vue
dantownsend: It's a good idea - the page is a bit busy at the moment.

We can't emphasis the response message too much, in case someone returns something really long (for example, some JSON).

Issue #173, opened 2022-07-04
Add translations


dantownsend:
Originally discussed here: https://github.com/piccolo-orm/piccolo/issues/556

We will have a new file called `translations.py` which will contain dictionaries mapping the English word to its translation. For example:

```python
FRENCH = {
    'table: '...',
    'submit': '...',
    ...
}
```

As a starting point, these are the translations we need most:

* Home
* Tables
* Log out
* Light Mode
* Dark Mode
* About
* Add Row
* Sort
* Show filters
* Export CSV
* Back
* Save
* Show referencing tables
* Hide referencing tables
* Delete
* Filter
* Close
* Apply
* Clear filters
* Per page
* Select one of the tables in the sidebar to get started.
* Language

We will add two new endpoints to the API - one for listing all available languages, and the other for retrieving a specific language (which is just the relevant dictionary as JSON).

```
GET /languages/
{'languages': ['French', ...]}

GET /languages/french/
{'translations': {'tables': '...'}}
```

The translations can be stored in the Vuex store.

The UI could look something like this:

![piccolo_admin_translations](https://user-images.githubusercontent.com/350976/177139105-e01512b0-6bc4-47e5-b128-c1f2588ab3a4.png)

We will add a global filter in Vue JS. So in the Vue code, we can do something like:

```
<h1>{{ 'Tables' | translate }}</h1>
```

sinisaos:
@dantownsend One question. If the translations are only going to be used in Piccolo Admin, why don't we use the standard tool for this in Vue, which is the [Vue I18n](https://kazupon.github.io/vue-i18n/). We just need to add the translation to e.g. `german.json` (same as we need to add translated words to `translations.py`) and everything is ready for internationalization. With Vue I18n we don't need to write our own template filters to map the translated words from `localStorage` to the template because Vue I18n has that built in (`{{ $t('About') }}` will be translated).

dantownsend:
@sinisaos I haven't used that library, but if it works, and it's not too large of a dependency, then I'm happy to use it.

sinisaos:
@dantownsend Today I tried this library for the first time when I saw this translation issue and it is easy to use. Bundle size is [here](https://bundlephobia.com/package/vue-i18n@8.27.1).

sinisaos:
@dantownsend If you want, I can do a PR tomorrow so you can see what it's all about (the library is easy to use and everything works great). I would also add a Croatian translation.

dantownsend:
[...]

PR #175, opened 2022-07-07
add translations

sinisaos:
Related to #173.

Result is:

https://user-images.githubusercontent.com/30960668/177712285-56767f52-468d-434e-a6b4-39e95edf6329.mp4

To add a new translation, we just need to add a new json (eg. `german.json`) to `assets/locales` and update `main.ts` to register the new translation.


dantownsend:
This looks cool. I haven't had time to properly review yet, so will put it in the next release.

There seems to be some merge conflicts as we've updated master.

sinisaos:
@dantownsend I merged the conflicts and push commit to update the translation for the new code from master

dantownsend:
@sinisaos Thanks

dantownsend:
Is there some way we can let the user specify a default language to use? For example, if the developer knows their users speak French, then it would be good to set that as a default?

Perhaps we can modify the meta endpoint, so it includes default language?

Also, I wonder if there's a way of getting VueI18n to lazy load languages, only when the user selects it from the dropdown menu? Otherwise, if we have lots of languages in the future, the bundle size will get much larger.

For example, when a language is clicked on in the drop down menu, axios is used to fetch the JSON, and register it with VueI18n.

sinisaos:
@dantownsend Good suggestions. I'll see what I can do. I also have to push a commit because I made a typo in the `assets` folder name (I wrote `assests`).

dantownsend:
@sinisaos Cool, thanks!

sinisaos:
@dantownsend I added a default language option to `create_admin`. Usage is:
```python
admin = create_admin(..., default_language="welsh")
```
When user login in admin `default_language` is loaded, and If we doesn't set `default_language`, default value is `"english"`.

There is an option for lazy loading languages, but from `locale` folder `lang.json`. I found something on Stackoverflow to load the translation from the api, but I haven't tried it yet.

dantownsend:
@sinisaos Great, thanks! I'll give it a go today.

sinisaos:
@dantownsend I managed to use the translations from the api by saving the api calls to `localStorage` (none of the ways I found online worked so I solved it this way). Now, when changing the language, we use data from `localStorage` and no longer from the `locale` folder. We now add translations to `translations.py`. Registration and usage of translations in Vue is now automatic.

admin_ui/src/components/NavTranslationsMenu.vue
```diff
@@ -0,0 +1,46 @@
...
+
+<script lang="ts">
+import Vue from "vue"
+import axios from "axios"
+import DropDownMenu from "./DropDownMenu.vue"
+
+export default Vue.extend({
+    props: {
+        showLanguageDropdown: Boolean
+    },
```
dantownsend, line 26: I don't think we're using this prop any more.

admin_ui/src/components/NavTranslationsMenu.vue
```diff
@@ -0,0 +1,46 @@
...
+    components: {
+        DropDownMenu
+    },
+    methods: {
+        closeDropdown() {
+            this.$emit("closeDropdown", false)
+        },
+        async getTranslation(value) {
+            const response = await axios.get(`./api/languages/${value}`)
```
dantownsend, line 35: Tiny nit-pick - add a trailing slash to the URL for consistency.

admin_ui/src/views/Login.vue
```diff
@@ -68,8 +63,11 @@ export default {
             } else {
                 this.$router.push({ name: "home" })
             }
-        },
+        }
     },
+    updated() {
+        this.$i18n.locale = this.defaultLanguage
+    }
```
dantownsend, line 70: Do we need this? Can we do it all in App.vue.

``updated`` gets called every time the DOM updates.

admin_ui/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "allowSyntheticDefaultImports": true,
     "sourceMap": true,
     "baseUrl": ".",
+    "resolveJsonModule": true,
```
dantownsend, line 14: We can probably remove this now, as we're not using the JSON files.

docs/source/internationalization/index.rst
```diff
@@ -0,0 +1,30 @@
...
+
+
+Usage
+-----
+
+From the language dropdown menu choose the language for which you want a translation. 
+
+
+.. hint::
+    More languages ​​will be supported in the future.
```
dantownsend, line 30: Maybe we can change this to:

```
More languages ​​will be supported in the future. Pull requests are welcome.
```

piccolo_admin/endpoints.py
```diff
@@ -558,6 +577,32 @@ def get_table_list(self, request: Request) -> t.List[str]:
...
+                {"error": "Language does not exist."}, status_code=422
+            )
+
+    def get_translations_list(
+        self, request: Request
+    ) -> t.Dict[str, t.Dict[str, str]]:
+        """
+        Returns a list of all languages.
+        """
+        return {i["language"]: i for i in TRANSLATIONS}
```
dantownsend, line 604: If `TRANSLATIONS` was a dict, we could do `list(TRANSLATIONS.keys())`.

admin_ui/src/components/NavTranslationsMenu.vue
dantownsend: `async getTranslation(value: string)` might be better to keep TypeScript happy

sinisaos: You're right. I forgot to delete this method. We don't use that. We don't even use the `get_translation` endpoint, but I left it in (I can delete it if you want).

piccolo_admin/endpoints.py
```diff
@@ -558,6 +577,32 @@ def get_table_list(self, request: Request) -> t.List[str]:
...
+    ###########################################################################
+
+    def get_translation(
+        self, request: Request, language: str
+    ) -> t.Union[JSONResponse, t.Dict[str, str]]:
+        """
+        Return a single language.
+        """
+        try:
+            language = request.path_params.get("language", None)
```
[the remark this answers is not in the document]
sinisaos: You're right, I will delete that. We don't use the `get_translation` endpoint, but I left it in (I can delete it if you want).

piccolo_admin/endpoints.py
sinisaos: With dictionary it will be as simple as `return TRANSLATIONS`.

admin_ui/src/views/Login.vue
sinisaos: Unfortunately, from `created()` in `App.vue` doesn't work. It only works from `Login.vue` (when admin is login to UI `default_language` is set). I'll change to `mounted()` lifecycle hook.

sinisaos: I left the` updated()` hook because if I use `mounted()` and, for example, refresh the login page, the default language is not applied in the UI. If you have any other idea, feel free to change it.

admin_ui/src/App.vue
```diff
@@ -121,6 +120,14 @@ export default Vue.extend({
             }
         )
 
+        axios.interceptors.request.use((config) => {
+            const languages = JSON.parse(localStorage.getItem("languages"))
+            for (const lang in languages) {
+                this.$i18n.setLocaleMessage(lang, languages[lang])
+            }
+            return config
+        })
```
dantownsend, line 129: I'm not sure about this - it will trigger on every axios request. Better just to fetch the languages and then store them once.

dantownsend:
Closing, as these changes were merged in another PR

Issue #172, opened 2022-07-03
WYSIWYG for admin


tigerline86:
Hello!
Could you tell me please, how i make override rich_text_columns
I want to use TinyMCE editor for Text columns!

sinisaos:
@tigerline86 In Piccolo Admin, we use the [vue2-editor](https://www.npmjs.com/package/vue2-editor) that uses Quill under the hood (among the most popular WYSIWYG editors for Vue). You can't override the vue2-editor, but you can integrate TinyMCE yourself (or another editor because the content is saved [here](https://github.com/piccolo-orm/piccolo_admin/blob/2fbdec3a63c919bc4c140b75ec5469b4921e78d0/admin_ui/src/components/InputField.vue#L65-L70) regardless of the WYSIWYG editor). Here is the changes you need to make in `InputField.vue`. 
```Vue
<template>
     ... // unchanged code
    <div v-else-if="format == 'text-area' && isFilter == false">
      <textarea
          v-if="
              !schema.rich_text_columns.includes(
                  getFieldName(title).toLowerCase()
              )
          "
          autocomplete="off"
          ref="textarea"
          v-bind:name="getFieldName(title)"
          v-bind:placeholder="placeholder"
          v-bind:style="{ height: textareaHeight }"
          v-model="localValue"
          v-on:input="setTextareaHeight"
      />
    
      <editor
          v-else
          v-model="localValue"
          api-key="no-api-key"
          :init="{
              height: 500,
              menubar: false,
              plugins: [
                  'advlist autolink lists link image charmap print preview anchor',
                  'searchreplace visualblocks code fullscreen',
                  'insertdatetime media table paste code help wordcount'
              ],
              toolbar:
                  'undo redo | formatselect | bold italic backcolor | \
                  alignleft aligncenter alignright alignjustify | \
                  bullist numlist outdent indent | removeformat | help'
          }"
      />
    
      <textarea
          id="editor"
          v-model.lazy="localValue"
          v-bind:name="getFieldName(title)"
      ></textarea>
    </div>

    ... // unchanged code

</template>

<script lang="ts">
import Vue, { PropType } from "vue"

import flatPickr from "vue-flatpickr-component"

import ArrayWidget from "./ArrayWidget.vue"
import ChoiceSelect from "./ChoiceSelect.vue"
import DurationWidget from "./DurationWidget.vue"
import OperatorField from "./OperatorField.vue"
import { Choices } from "../interfaces"
[...]

PR #176, opened 2022-07-07
headings option to WYSIWYG

sinisaos:
Related to #172

### DIFF
--- a/admin_ui/src/components/InputField.vue
+++ b/admin_ui/src/components/InputField.vue
@@ -223,7 +223,7 @@ export default Vue.extend({
                     { align: "justify" }
                 ],
                 ["link", "image", "code-block"],
-                [{ header: [1, 2, 3, 4, 5, 6] }]
+                [{ header: [false, 1, 2, 3] }]
             ]
         }
     },


dantownsend:
Screenshot 2022-07-07 at 22 08 38

Can we add a way of turning the heading back to a non-heading?

sinisaos:
@dantownsend Unfortunately, this is not a default option, in fact there is no such option in Quill at all. I found one [gist](https://gist.github.com/izzygld/0762dce7292baf2168f9c7b2aa99b535) that I tried and modified, but I couldn't get the `showHtml` button to appear. I pushed a new commit to add an option to turning the heading back to a non-heading.

I think we went a bit too far with this (based on issue raised about WYSIWYG editor). Vue2-editor (Quill) is a good editor with enough functions. Rich text fields are just an option in Piccolo Admin that the user may or may not use at all. Even if that's not enough, we are intentionally made that content is saved from [here](https://github.com/piccolo-orm/piccolo_admin/blob/master/admin_ui/src/components/InputField.vue#L65-L69) and anyone can implement a WYSIWYG (not ideal because the user needs to fork the codebase, but there's still that option and an example of how to do that) editor they want, like I wrote in issue. I think that is more than enough from our side.

Issue #181, opened 2022-07-12
Image upload


tigerline86:
Hello!
Please tell me if there are any built-in mechanisms for uploading images in the admin panel?

sinisaos:
@tigerline86 For now, there is no such possibility. There is already a issue #109 about that. I was able to upload images from the frontend using the Cloudinary widget (there is video about that), but we think that upload from the backend would be a better option because that way the user can write their own adapter to work with any storage provider they want (including  local storage).

dantownsend:
Yeah, as @sinisaos says, we're working on it at the moment.

One hack is to use a rich text field, and add images there.

sinisaos:
@dantownsend That's right. Quill is very good for this because you don't have to do any extra configuration to upload images. Everything simply works out of the box.

dantownsend:
This has been implemented now:

https://piccolo-admin.readthedocs.io/en/latest/media_storage/index.html

Issue #109, opened 2021-10-27
`TableConfig` media handlers


dantownsend:
This is a big task, but unlocks a lot of great functionality in the admin.

Piccolo doesn't have a particular column type for storing media (images, video etc). It's possible to store them in binary columns, but this is generally frowned upon as it quickly leads to database bloat (and database storage is typically much more expensive than object storage).

The usual approach for storing media is to put the file in object storage (like S3), or in a local directory, and store a reference to it in the database (for example a URL or path).

Initially we can just store the media files in a local directory, but supporting the S3 protocol would be great in the future.

The steps required to implement this:

## MediaHandler class

```python
# This is just an abstract base class for all other media handlers:
class MediaHandler:
    pass

@dataclass
class LocalMediaHandler(MediaHandler):
    root_path: str

    def store_file(self, file):
        ...

    def retrieve_file(self, path: str):
        ...
```

## TableConfig changes

```python
media_handler = LocalMediaHandler(root_path='/var/www/files/')

app = create_admin(
    TableConfig(
        Movie,
        media_handler=media_handler,
        media_columns=[Movie.poster]
    )
)
```

## PiccoloCRUD changes

There will be a new endpoint, something like:

```
/tables/director/1/media?column=poster
```

Which can be used to retrieve the actual media file.

## UI improvements

The UI will know if a column is used for media storage by inspecting the JSON schema:

```
/api/tables/director/schema
```

`PiccoloCRUD` would've added an extra field to it called `'media_columns'`, using `create_pydantic_model`'s `schema_extra_kwargs` feature.

The UI would then render a file upload box for that field. In the future we could support simple editing functionality like cropping images (there are several JS libraries for this), but that's down the road.

sinisaos:
@dantownsend I did a little research on the internet and this proposed feature is quite different from uploading from html forms. Upload files is not a problem, we can do it like this to upload files to local static folder.
```python
# in endpoints.py
api_app.add_api_route(
  path="/media/",
  endpoint=self.upload_images,  # type: ignore
  methods=["POST"],
  tags=["Media"],
)
...
def upload_images(self, files: t.List[UploadFile] = File(...)):
  for f in files:
    with open(
      f"static/{random.randint(100,100000)}.jpeg", "wb"
    ) as buffer:
      shutil.copyfileobj(f.file, buffer)
```
Result is:

[...]

PR #182, opened 2022-07-13
upload images to local storage

sinisaos:
Related to #181 and #109. For now, the upload is only in a local static folder.


sinisaos:
@dantownsend When you have time can you check if this example of S3 adapter for media upload works. It should work.
```python
# in example.py
import aiobotocore

@dataclass
class S3MediaHandler:
    aws_access_key_id: str
    aws_secret_access_key: str
    region: str
    bucket: str
    key: str

    async def upload(self, request, file):
        session = aiobotocore.get_session()
        async with session.create_client(
            "s3",
            region_name=self.region,
            aws_secret_access_key=self.aws_secret_access_key,
            aws_access_key_id=self.aws_access_key_id,
        ) as client:
            await client.put_object(
                ACL="public-read", Bucket=self.bucket, Key=self.key, Body=file
            )
            return {
                "image": (
                    f"https://{self.bucket}.s3.{self.region}.amazonaws.com/{self.key}"
                ),
            }

media_handler: t.Any = S3MediaHandler(...)
```

dantownsend:
We need to be careful with this:

```python
            await client.put_object(
                ACL="public-read", Bucket=self.bucket, Key=self.key, Body=file
            )
            return {
                "image": (
                    f"https://{self.bucket}.s3.{self.region}.amazonaws.com/{self.key}"
                ),
            }
```

The URL is hardcoded to AWS, but it should be possible to use boto with any S3 compatible storage. I would just return the key, rather than the full URL.

We then need an endpoint which gets the URL for the image - for S3 it needs to create a signed URL. If we just store the image in a public bucket, it won't be secure. Something like:

```
GET /api/media/some-file-key

{
    'url': 'https://some-signed-url.com/foo/'
}
```

So in Piccolo Admin, when we load a column which has media storage enabled, we get the key for the image, and then fetch a signed URL on demand for that image.

You get a signed URL doing something like this (I lifted this code from one of my projects):

```python
response = s3_client.generate_presigned_url(
            ClientMethod="get_object",
            Params={"Bucket": BUCKET_NAME, "Key": image_name},
            ExpiresIn=3600,
        )
```

I hope that makes sense.

sinisaos:
@dantownsend I agree with you and it should be changed, but this is just an example of a media handler that the user should have written themselves (as we said before, user can write their own adapter to work with whatever storage provider they want). I'm mostly interested in whether that handler works with the `store_files()` method from `endpoints.py`.

dantownsend:
I see. Yeah, it looks like it should work.

Have you ever tried this?

http://docs.getmoto.org/en/latest/docs/server_mode.html#run-using-docker
https://hub.docker.com/r/motoserver/moto/tags

It's a Docker container you can run which acts as a fake S3.

sinisaos:
@dantownsend Sorry for the slow response but I was unable to test the `/api/media/` endpoint. Thanks for the suggestions. I'll try tomorrow.

sinisaos:
@dantownsend I changed the adapter to this one.
```python
import boto3
from botocore.exceptions import ClientError

@dataclass
class S3MediaHandler:
    aws_access_key_id: str
    aws_secret_access_key: str
    bucket: str
    key: str
    region: str

    def upload(self, request, file):
        s3_client = boto3.resource(
            "s3",
            aws_access_key_id=self.aws_access_key_id,
            aws_secret_access_key=self.aws_secret_access_key,
            region_name=self.region,
        )
        s3_client.create_bucket(Bucket=self.bucket)
        # Upload the file
        s3_client = boto3.client("s3")
        try:
            response = s3_client.upload_fileobj(
                file.file,
                self.bucket,
                self.key,
            )
        except ClientError as e:
            print(e)  # or logging

        response = s3_client.generate_presigned_url(
            ClientMethod="get_object",
            Params={"Bucket": self.bucket, "Key": file.filename},
            ExpiresIn=3600,
        )

        return {"image": response}

media_handler: t.Any = S3MediaHandler(...)
```
I hope this works, but you'll have to try it yourself. To be honest, I've been struggling with the moto library (and I don't know if I've used it well at all) and I've used the `@mock_s3` decorator to test the `/api/media/` endpoint with the moto server, but I get `NoSuchKey The specified key does not exist. 1234` (`1234` is the image name) when I open the response url in the browser.

dantownsend:
@sinisaos I really struggled using moto at first. It took me ages to work it out.

I hoped that using the Docker container gave a simpler experience, and you could just point your code at it, and it would act like a local version of S3.

Don't worry too much about S3 for now - we can always add it later. Just local storage will be super useful.

sinisaos:
@dantownsend Ok. Upload to local storage (code is in this PR) is ready and everything works well. If you have time, can you try it?

piccolo_admin/endpoints.py
```diff
@@ -486,6 +505,11 @@ def __init__(
             app=StaticFiles(directory=os.path.join(ASSET_PATH, "js")),
         )
 
+        self.mount(
+            path="/static",
```
dantownsend, line 509: I would mount this at `/media/` instead - usually static refers to an application's own files, and media refers to uploaded files.

docs/source/media_upload/index.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Piccolo Admin has the option of uploading media files to local static folder 
+(local storage). Multi file upload is enabled per record for all columns 
+specified in ``media_columns``. 
+
+
+Usage
+-----
+
+Piccolo admin uses the Piccolo ORM `Array <https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/column_types.html#array>`_
```
dantownsend, line 16: I'm not 100% sure about it having to be an `Array` column. I can imagine situations where someone wants an array, but I can also imagine situations where someone wants to use a `Varchar`, as they only want a single image to be uploaded.

piccolo_admin/example.py
```diff
@@ -180,6 +189,22 @@ async def booking_endpoint(request, data):
     return "Booking complete"
 
 
+@dataclass
+class LocalMediaHandler:
```
dantownsend, line 193: I would make this a standard Python class, and have it inherit from a base class, called something like `MediaHandler`.

```
class MediaHandler(abc.ABC):
    @abc.abstract_method
    def upload(self, request, file):
        pass
```

Then it's more obvious what a user needs to implement if they create their own handler.

dantownsend:
This looks cool. I've got quite a few ideas about this - do you want me to make a branch and have a play around?

docs/source/media_upload/index.rst
sinisaos: The input type file is set to upload only one file, so the user can easily upload only one image. For uploading multiple files `Array` column is ideal (we already have a `ArrayWidget` for adding and removing fields which is very useful), so this solves the uploading of multiple or single file in one solution.

piccolo_admin/endpoints.py
sinisaos: You're right, I'll change that.

docs/source/media_upload/index.rst
dantownsend: Yeah, in some situations though you want to make sure the user can only upload one file.

Imagine you're building a user profile table, and you have a field called 'profile_image'. If there are multiple images stored, it will be confusing - which is the genuine one?

sinisaos:
Of course. That would be great

docs/source/media_upload/index.rst
sinisaos: I made 2 videos to see what I mean.
Single file upload

https://user-images.githubusercontent.com/30960668/180474565-79feda60-3c0d-4bc0-9ebe-6a6e3428f882.mp4

Multiple files upload


https://user-images.githubusercontent.com/30960668/180474867-9a63121c-bb8e-41b2-95df-0a3d27f30994.mp4

But of course you can do it however you want, but I think this is a simple and efficient solution because we actually upload only one file each time.

dantownsend:
Changes were merged in via this PR: https://github.com/piccolo-orm/piccolo_admin/pull/189

PR #183, opened 2022-07-19
Translation experiment

dantownsend:
Based on https://github.com/piccolo-orm/piccolo_admin/pull/175

I added a few things:

 * Some extra languages
 * A script (`get_translations.py`) for checking we have all of the required translations
 * Using language codes to refer to languages - so we can add different dialects in the future (e.g. American English vs British English)

There's a few minor things left to do:

 * The localStorage isn't quite working correctly - when the user loads the page, if they previously set their language preference, it should automatically use it. Shouldn't take long.
 * Possibly integrate `navigator.language` - so we can select the user's preferred language without them having to do anything.
 * Might move the endpoint for fetching translations , so it doesn't require auth - so we can call it from the login page.


piccolo_admin/endpoints.py
```diff
@@ -658,6 +722,48 @@ def create_admin(
...
+                tables=[TableA, TableB],
+                default_language_code='hr',
+                translations=[ENGLISH, CROATIAN]
+            )
+
+        You can also use this to provide your own translations, if there's a
+        language we don't currently support (though please open a PR to add
+        it!)::
+
+            from piccolo.translations import Translation, ENGLISH
```
sinisaos, line 750: I think it should be for imports
```python
from piccolo_admin.translations.data import ENGLISH
from piccolo_admin.translations.models import Translation
```

piccolo_admin/endpoints.py
```diff
@@ -658,6 +722,48 @@ def create_admin(
...
+        will be automatically translated into this language.
+    :param translations:
+        Specify which translations are available. By default, we use every
+        translation in ``piccolo_admin/translations/data.py``.
+
+        Here's an example - if we know our users only speak English or
+        Croatian, we can specify that only those translations are visible
+        in the language selector in the UI::
+
+            from piccolo.translations import ENGLISH, CROATIAN
```
sinisaos, line 738: I think it should be for imports
```python
from piccolo_admin.translations.data import ENGLISH
from piccolo_admin.translations.models import Translation
```

piccolo_admin/endpoints.py
```diff
@@ -560,12 +592,42 @@ def get_meta(self) -> MetaResponseModel:
...
+        Return a list of language codes and names for each available
+        translation.
+        """
+        return TranslationListResponse(
+            translations=[
+                TranslationListItem(
+                    language_code=translation.language_code,
+                    language_name=translation.language_name,
+                )
+                for translation in TRANSLATIONS
```
sinisaos, line 614: This should be `for translation in self.translations_map.values()` because if we set only English and Croatian in `create_admin(translations=[ENGLISH, CROATIAN])` all translations are in language dropdown and that is not right because languages ​​that are not set in `create_admin` do not work and return `404 Page not found` error.

dantownsend: Good catch!

admin_ui/src/modules/translations.ts
```diff
@@ -0,0 +1,84 @@
...
+        /**
+         * If the user has previously used the app, and set their default
+         * language, then load it from localStorage.
+         */
+        async loadDefaultLanguage(context) {
+            const defaultLanguage = localStorageUtils.getDefaultLanguage()
+            if (defaultLanguage) {
+                context.commit("updateDefaultLanguage", defaultLanguage)
+            }
+        }
```
sinisaos, line 82: This is currently not working. I don't know what you planned, but you can set `defaultLanguage` in `BaseView.vue`. First you need to add `localStorageUtils.setDefaultLanguage(response.data.default_language_code)` in `fetchTranslations` (set `defaultLanguage` to `localStorage` because it is `null` now) and after that `BaseView.vue` can look like this:
```javascript
<script lang="ts">
import Vue from "vue"
import axios from "axios"
import NavBar from "../components/NavBar.vue"
import SidebarNav from "../components/SidebarNav.vue"
import i18n from "../i18n"
import { TranslationAPIResponse } from "../interfaces"

export default Vue.extend({
    components: {
        NavBar,
        SidebarNav
    },
    computed: {
        username() {
            const user = this.$store.state.user
            return user ? user.username : "Unknown"
        },
        siteName() {
            return this.$store.state.metaModule.siteName
        }
    },
    async mounted() {
        const defaultLanguage = localStorage.getItem(
            "piccoloAdminDefaultLanguage"
        )
        const response = await axios.get<TranslationAPIResponse>(
            `./api/translations/${defaultLanguage}/`
        )
        i18n.setLocaleMessage(
            response.data.language_code,
            response.data.translations
        )
        i18n.locale = response.data.language_code
    }
})
</script>
```
With this we have a `defaultLanguage` persisted even if we refresh the page

piccolo_admin/endpoints.py
dantownsend: Thanks - will change it.

piccolo_admin/endpoints.py
dantownsend: Thanks - will change.

piccolo_admin/endpoints.py
sinisaos: I think you should change this because `piccolo.translations` module does not exist

admin_ui/src/modules/translations.ts
dantownsend: You're right - it still needs some work. I'll try what you suggested.

piccolo_admin/endpoints.py
sinisaos: I'm sorry. I didn't see you change it in the last commit.

admin_ui/src/modules/translations.ts
sinisaos: This also doesn't work well.

dantownsend: I haven't had a chance yet.

dantownsend: I gave it a go, and I think everything is working now.

Will you try the latest changes?

One thing to be aware of is I changed the URLs slightly. If you're using Nginx locally as a proxy, it needs updating to this:

https://github.com/piccolo-orm/piccolo_admin/blob/7871fd5d43ea21d505d7d7cbf725b168011277b1/extra/piccolo_admin

Now all of the public endpoints are under `/public/` and all of the private endpoints are under `/api/`. It makes things a bit cleaner.

piccolo_admin/endpoints.py
```diff
@@ -262,6 +268,8 @@ def __init__(
         rate_limit_provider: t.Optional[RateLimitProvider] = None,
         production: bool = False,
         site_name: str = "Piccolo Admin",
+        default_language_code: str = "auto",
```
sinisaos, line 271: It should revert to `default_language_code: str = "en"` because it is English is default language

piccolo_admin/endpoints.py
```diff
@@ -612,6 +670,8 @@ def create_admin(
     rate_limit_provider: t.Optional[RateLimitProvider] = None,
     production: bool = False,
     site_name: str = "Piccolo Admin",
+    default_language_code: str = "auto",
```
sinisaos, line 673: It should revert to `default_language_code: str = "en"` because it is the English is default language

admin_ui/src/modules/translations.ts
```diff
@@ -0,0 +1,107 @@
...
+            // If the user has previously used the app, and set their default
+            // language, then load it from localStorage:
+            const storedLanguagePreference =
+                localStorageUtils.getDefaultLanguage()
+
+            if (
+                storedLanguagePreference &&
+                availableLanguageCodes.indexOf(storedLanguagePreference) != -1
+            ) {
+                defaultLanguage = storedLanguagePreference
```
sinisaos, line 62: This should be `defaultLanguage = response.data.default_language_code` because otherwise we get the last selected language from the dropdown as the default, which is not the case. We should get the default language as we specified in `create_admin()`. That's why we need `default_language_code: str = "en"` because if it is set as `auto` during logout, we get a `404` error because `http://piccolo_admin/translations/auto` does not exist.

sinisaos:
@dantownsend I left a few comments that I think should be changed. You should also add `__init__.py` to translations module (otherwise we get `ModuleNotFoundError`). Everything else looks and works great. Grate job.

tests/test_endpoints.py
```diff
@@ -383,6 +383,32 @@ def test_get_user(self):
...
+class TestTranslations(TestCase):
+    def test_translations(self):
+        client = TestClient(APP)
+
+        response = client.get("/public/translations/")
+        self.assertEqual(response.status_code, 200)
+
+        data = response.json()
+
+        self.assertEqual(data["default_language_code"], "auto")
```
sinisaos, line 395: If we set `default_language_code` to `"en"` we need to change this to `self.assertEqual(data["default_language_code"], "en")`

admin_ui/src/modules/meta.ts
```diff
@@ -1,23 +1,28 @@
...
+import axios from "axios"
 
 export default {
     state: {
-        siteName: 'Piccolo Admin',
-        piccoloAdminVersion: 'Unknown'
+        siteName: "Piccolo Admin",
+        piccoloAdminVersion: "Unknown",
+        defaultLanguage: "english",
+        languages: undefined
```
sinisaos, line 8: I think we can remove this as it is not being used.

admin_ui/src/modules/translations.ts
dantownsend: Yeah, I intended it to the be like that. So the logic is:

 * If the default language is set to 'auto', get the user's language preference from the browser. If we don't have translations available for their language, it falls back to English.
 * If the default language is something other than 'auto', use that language by default.
 * When the user picks a different language from the dropdown, that becomes their default.

If the app is makin an API call to `/translations/auto` then that's definitely a bug though.

sinisaos: I'm sorry I didn't realize that. Then forget my comments about it. Everything is fine and there are no bug because `/translations/auto` is called only if we change the `default_language_code` to `"en"`. If it is `"auto"` everything is fine.

dantownsend: No worries - the logic changed quite a bit. It's more complex, but I think it covers all use cases.

admin_ui/src/modules/meta.ts
dantownsend: Good catch - will remove them!

admin_ui/src/views/Login.vue
```diff
@@ -3,77 +3,67 @@
...
+            password: ""
         }
     },
     computed: {
         siteName() {
             return this.$store.state.metaModule.siteName
         },
+        defaultLanguage() {
+            return this.$store.state.metaModule.defaultLanguage
+        }
```
sinisaos, line 34: I think we can remove this as well since it's not being used.

dantownsend: You're right - thanks. Have removed it now.

sinisaos: Great. Can you please add `__init__.py` to the translation module (otherwise we get `ModuleNotFoundError`). The docs should also be updated with the new features and it would be great if you could add a new image as the drop down now looks a lot better.

dantownsend:
> @dantownsend I left a few comments that I think should be changed. You should also add `__init__.py` to translations module (otherwise we get `ModuleNotFoundError`). The docs should also be updated with the new features and it would be great if you could add a new image as the dropdown now looks a lot better.
> 
> Everything else looks and works great. Great job.

Cool, thanks. I've updated the docs, and added the `__init__.py`.

sinisaos (approved):
I think this is great and you should merge this.

dantownsend:
@sinisaos Thanks, will merge it now. 👍 

Ended up being a big feature, but I think it's a useful one.

PR #188, opened 2022-07-29
Add Ukrainian translation for GUI

ruslan-rv-ua:


dantownsend:
@ruslan-rv-ua Perfect, thanks! 👍

PR #189, opened 2022-07-30
File upload support

dantownsend:
Based on https://github.com/piccolo-orm/piccolo_admin/pull/182

@sinisaos I've done a bunch of work on this.

The main changes are:

 * Added extra security based on this [OWASP doc](https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html)
 * It now supports all kinds of files - e.g. video as well
 * Added a media preview window, so the user can see the full sized image or video within Piccolo Admin
 * `Array`, `Varchar`, and `Text` columns can all be used as media columns

You can see the changes here:

https://user-images.githubusercontent.com/350976/181996743-17dae1d0-8b13-42de-89c0-b16c5643997b.mp4

 The biggest change from the original PR is it saves a unique identifier for the file in the database, rather than the URL. There are two reasons for this:

 * If the user decides to host their media files somewhere else, they don't have to update anything in the database.
 * It's the only way to get it working with S3. Each time someone accesses a file in S3 we need to generate a signed URL, which is only valid for a short period of time, so we can't store the URL in the database.

I think it's about 80% done. It's a massive feature, but a really cool one.

Things that need doing still:

- [x] Update the docs
- [ ] I need to add your file upload button back - it looks nicer than what's there now
- [x] Add links on the row detail page for viewing images
- [x] Maybe add S3 support, so we know the API works properly with it before releasing it
- [x] The media is always served from `/api/media-files/` at the moment. It's ignoring the `media_url` setting in `LocalStorage`. That needs some thinking about.
- [x] We used to have `media_columns` and `media_storage` arguments in `TableConfig`. I changed it slightly (now there's just `media_columns`, which maps a column to a `MediaStorage`) because I can imagine use cases where you want some columns to store files in one place, and other columns to store files in another place. Also, you might only allow images in some columns, and videos in other columns. It still needs a bit of thinking about.
- [ ] I might move a bunch of the logic to Piccolo API, so it can also be used with `PiccoloCRUD` standalone.

I removed your image previews from `RowListing.vue` for now. I liked the way they looked, but because we're not creating thumbnails for the images, if the page is showing 100 rows, it will pull down 100 full sizes images, which would slow things down too much. Once we work out thumbnails I would like to add it back.


admin_ui/src/components/MediaViewer.vue
```diff
@@ -0,0 +1,213 @@
...
+        <div class="bottom_bar">
+            <p>
+                <a class="button" :href="fileURL" download
+                    ><font-awesome-icon icon="download"></font-awesome-icon>
+                    Download</a
+                >
+            </p>
+        </div>
+    </div>
+</template>
```
sinisaos, line 50: This is great. I really like it. Looks cool.

piccolo_admin/endpoints.py
```diff
@@ -416,6 +460,32 @@ def __init__(
...
+            response_model=GenerateFileURLResponseModel,
+        )
+
+        # TODO - this is problematic ... mount a media folder by default?
+        # Allow the user to customise the location???
+        private_app.mount(
+            path="/media-files/",
+            app=StaticFiles(directory=MEDIA_PATH),
+        )
+
```
sinisaos, line 488: I don't think this is too much of a problem since even Django have the folder for media files under MEDIA_ROOT. Maybe we can just rename `example_media` to `uploads` to make it clearer what the purpose of that folder is.

sinisaos:
> Things that need doing still:
> 
> * Update the docs
> * I need to add your file upload button back - it looks nicer than what's there now

I agree. Upload button looks much better.

> * Add links on the row detail page for viewing images
> * Maybe add S3 support, so we know the API works properly with it before releasing it
> * The media is always served from `/api/media-files/` at the moment. It's ignoring the `media_url` setting in `LocalStorage`. That needs some thinking about.

I think we should put a default `media_url` (which cannot be changed by the user) but as you wish. Both ways are good.

> * We used to have `media_columns` and `media_storage` arguments in `TableConfig`. I changed it slightly (now there's just `media_columns`, which maps a column to a `MediaStorage`) because I can imagine use cases where you want some columns to store files in one place, and other columns to store files in another place. Also, you might only allow images in some columns, and videos in other columns. It still needs a bit of thinking about.
> * I might move a bunch of the logic to Piccolo API, so it can also be used with `PiccoloCRUD` standalone.

That would be very nice and useful for the end user, because uploading files is not an easy task, and with that, they would have everything prepared on the backend side and should only make the frontend (even they can use Piccolo Admin frontend as an example).

> 
> I removed your image previews from `RowListing.vue` for now. I liked the way they looked, but because we're not creating thumbnails for the images, if the page is showing 100 rows, it will pull down 100 full sizes images, which would slow things down too much. Once we work out thumbnails I would like to add it back.

That would be nice. Did you mean to use something like PIL on the backend or something on the frontend to generate the thumbnails on the fly?

And for the end, this is really great and this would be **killer new feature** which I think will attract a lot of new users.. It is more complex, but that's why it covers all possibilities (files names and extensions) and provides additional security.

dantownsend:
@sinisaos Thanks for the feedback. Will hopefully have it finished soon.

admin_ui/src/components/MediaViewer.vue
dantownsend: Thanks!

admin_ui/src/components/InputField.vue
```diff
@@ -1,5 +1,20 @@
 <template>
     <div>
+        <template v-if="isMediaColumn && !isFilter">
+            <div class="media_block">
+                <input type="file" @change="uploadFile($event)" />
+                <a href="#" @click.prevent="showMedia" v-if="localValue"
+                    >View</a
+                >
```
sinisaos, line 8: I don't know if this helps, but you can write this part like this  
```vue
<a
    href="#"
    @click.prevent="showMedia"
    v-if="localValue && type !== 'array'"
    ><font-awesome-icon icon="eye" size="lg" title="View" // nedd to add faEye to fontawesome.ts
/></a>
```
so that the `View` button is not displayed in the array widget.

admin_ui/src/components/ArrayWidget.vue
```diff
@@ -13,7 +13,12 @@
                 </a>
             </li>
             <li>
-                <a href="#" v-on:click.prevent="addArrayElement">
+                <a
+                    class="add"
+                    href="#"
+                    v-on:click.prevent="addArrayElement"
+                    v-if="enableAddButton"
+                >
```
sinisaos, line 21: I don't know if this helps, but you can write whole file like this because with that we can see every single file in array widget (I don't know if you have any other idea but it doesn't work at the moment)
```vue
<template>
    <div>
        <ul class="array_items">
            <li :key="index" v-for="(value, index) in internalArray">
                <input
                    :type="inputType"
                    :value="value"
                    id="choice"
                    v-on:change="updateArray($event, index)"
                />
                <a
                    href="#"
                    v-on:click.prevent="showMedia(index)"
                    :value="value"
                    v-if="isMediaColumn"
                    ><font-awesome-icon icon="eye" title="View"
                /></a>
                <a href="#" v-on:click.prevent="removeArrayElement(index)">
                    <font-awesome-icon icon="times" />
                </a>
            </li>
            <li>
                <a
                    class="add"
                    href="#"
                    v-on:click.prevent="addArrayElement"
                    v-if="enableAddButton"
                >
                    <font-awesome-icon icon="plus" />{{ $t("Add") }}
                </a>
            </li>
        </ul>
        <MediaViewer
            v-if="showMediaViewer"
            :mediaViewerConfig="mediaViewerConfig"
            @close="showMediaViewer = false"
        />
    </div>
</template>

<script lang="ts">
import MediaViewer from "./MediaViewer.vue"
import { MediaViewerConfig } from "@/interfaces"

export default {
    props: {
        array: {
            type: Array,
            default: () => []
        },
        inputType: {
            type: String,
            default: "text"
        },
        title: {
            type: String,
            default: ""
        },
        enableAddButton: {
            type: Boolean,
            default: true
        }
    },
    components: {
        MediaViewer
    },
    data() {
        return {
            internalArray: [],
            showMediaViewer: false,
            mediaViewerConfig: null as MediaViewerConfig,
            index: 0
        }
    },
    computed: {
        schema() {
            return this.$store.state.schema
        },
        currentTableName() {
            return this.$store.state.currentTableName
        },
        isMediaColumn() {
            return this.schema.media_columns.includes(
                this.getFieldName(this.title).toLowerCase()
            )
        }
    },
    methods: {
        updateArray($event, index) {
            this.$set(this.internalArray, index, $event.target.value)
            this.$emit("updateArray", this.internalArray)
        },
        addArrayElement() {
            this.internalArray = [...this.internalArray, ""]
            this.$emit("updateArray", this.internalArray)
        },
        removeArrayElement(index) {
            this.$delete(this.internalArray, index)
            this.$emit("updateArray", this.internalArray)
        },
        getFieldName(name: string) {
            return name.toLowerCase().split(" ").join("_")
        },
        showMedia(index) {
            const mediaViewerConfig: MediaViewerConfig = {
                fileKey: this.internalArray[index],
                columnName: this.getFieldName(this.title),
                tableName: this.currentTableName
            }
            this.mediaViewerConfig = mediaViewerConfig
            this.showMediaViewer = true
        }
    },
    watch: {
        array() {
            this.internalArray = [...this.array]
        }
    },
    mounted() {
        this.internalArray = [...this.array]
    }
}
</script>
```

dantownsend: You're right - the `view` button doesn't work for the array column at the moment.

What you've suggested here looks good.

sinisaos: Thanks. You can also remove the `static` folder as it is no longer used.

requirements/extra/s3.txt
```diff
@@ -0,0 +1 @@
+boto3==1.*
```
sinisaos, line 1: You should add `moto==3.*` as it is a missing requirement when I tried to run the tests.

dantownsend: I think I added it to `test-requirements.txt` - maybe I didn't commit it.

extra/piccolo_admin
```diff
@@ -2,6 +2,9 @@ server {
     listen 80;
     server_name piccolo_admin;
 
+    // By default Nginx only allows us to upload very small files.
+    client_max_body_size 20M;
```
sinisaos, line 6: You should add a `#` at the beginning of the comment or remove it, otherwise [nginx](http://nginx.org/en/docs/beginners_guide.html#conf_structure) failed to start.

requirements/extra/s3.txt
sinisaos: Sorry. You are right `moto` is in `test-requirements.txt`

extra/piccolo_admin
dantownsend: Good catch! Will fix it now.

docs/source/index.rst
```diff
@@ -1,25 +1,41 @@
...
+
+* Powerful data filtering
+* Builtin security
+* Media support, both locally and in S3 compatible services
+* Dark mode support
+* CSV exports
+* Easily create custom forms
+* Works on mobile and desktop
+* Use standalone, or integrate it easily with ASGI apps like FastAPI, and Starlette
+* Multilingual out of box
```
sinisaos, line 26: I think you should add `bulk data operations` and `simple table and column configuration` (how data is displayed -`TableConfig`), because those are two really nice features

sinisaos:
> My god, this is turning into a huge PR.

That's true, but you covered all the possible combinations  and obstacles (file names, file extensions, S3 upload, LocalMedia upload, deleting unused files etc. - I must stop here because list of features is really big :) ) I can only say **`BRAVO`**, because this is going to be really cool. I think with this Piccolo Admin will be the best and richest admin UI I have ever seen. The amount of built-in features is huge.

admin_ui/src/components/InputField.vue
sinisaos: This is a full template with an upload button that replaces the input file type and a view icon to display the file. I think that looks ok.
```vue
<template v-if="isMediaColumn && !isFilter">
    <div class="media_block">
        <form
            id="uploadForm"
            enctype="multipart/form-data"
            v-on:change="uploadFile($event)"
        >
            <label>
                Upload file
                <input type="file" name="file" />
            </label>
        </form>
    </div>

    <a
        href="#"
        @click.prevent="showMedia"
        v-if="localValue && type !== 'array'"
        ><font-awesome-icon icon="eye" size="lg" title="View"
    /></a>

    <MediaViewer
        v-if="showMediaViewer"
        :mediaViewerConfig="mediaViewerConfig"
        @close="showMediaViewer = false"
    />
</template>
...
<style>
div.media_block {
    display: inline-block;
}

input[type="file"] {
    display: none;
}

label {
    cursor: pointer;
    border: none;
    padding: 0.8rem 1.2rem;
    font-weight: bolder;
    margin-bottom: 0.5rem;
    transition: background-color 0.5s;
    background-color: @dark_blue;
    color: white;
    font-size: 0.7em;
    text-transform: uppercase;
    text-align: center;
}

a {
    float: right;
    margin-top: 0.7rem;
}
</style>
```
Result is
upload_button

docs/source/index.rst
dantownsend: Good idea - I'll add those.

dantownsend:
@sinisaos I'm going to merge this in now. It's not 100% perfect, but the rest can be picked up in future PRs.

sinisaos:
@dantownsend I think this is pretty good. Well done.

dantownsend:
@sinisaos Thanks

Issue #192, opened 2022-08-05
Move `MediaStorage` / `LocalMediaStorage` / `S3MediaStorage` to `piccolo_api`


dantownsend:
I would like to integrate our media storage code with `PiccoloCRUD` at some point. So really, the media storage code should live in Piccolo API (otherwise we will have circular dependency between `piccolo_admin` and `piccolo_api`).

So everything in `piccolo_admin/media` should move to `piccolo_api/media`. Likewise, with the associate tests.

The docs can stay as part of Piccolo Admin for now. We just need to update the imports so they're from `piccolo_api.media`.

dantownsend:
@sinisaos What do you think about this?

Just wondering whether to do it before releasing the next version of Piccolo Admin.

sinisaos:
@dantownsend I think it's a good idea because, as you said, we will prevent circular dependency, but as I mentioned before, it would be very nice and useful for the end user, because they would have everything prepared on the backend and only need to make the frontend.

sinisaos:
@dantownsend If you want I can make these changes in Piccolo API and Piccolo Admin? Of course, you'll have to check if I did everything right.

dantownsend:
@sinisaos If you wouldn't mind, that would be great - thanks

sinisaos:
@dantownsend I will do it during the day.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_admin/pull/193

PR #193, opened 2022-08-06
removed media files after switching to Piccolo API

sinisaos:
I hope you meant something similar to this. Related to #192 .


dantownsend:
@sinisaos That's great thanks - I'll just update the docs, and will merge it.

Issue #196, opened 2022-08-08
Make media upload endpoints obey `read_only` option


dantownsend:
Piccolo Admin can be used in `read_only` mode - it's mostly used for demos, and in situations where you really don't trust the end user.

We added media upload support, which means there are now two new endpoints.

The changes required are to these two methods:

https://github.com/piccolo-orm/piccolo_admin/blob/e94736b6135dc1d3b1b2563f5f62976cd244fdcc/piccolo_admin/endpoints.py#L648-L690

We need to return a 405 status code is if read only mode.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_admin/issues/196

PR #198, opened 2022-08-10
media upload endpoints obey read_only option

sinisaos:
Related to #196


dantownsend:
@sinisaos Thanks a lot for this - it's great.

I made two small changes:

 * I'm returning a custom message in the HTTP response, saying 'running in read-only mode'.
 * I changed the `if` statement slightly, so there's less indentation.

PR #202, opened 2022-08-13
page title, favicon and logo to UI

sinisaos:
Related to #201


sinisaos:
@dantownsend Now I understand why you said the favicon would be tricky. Test are failing because `icons` directory does not exists. How you usually test changes in public directory (`dist`)?

dantownsend:
@sinisaos No worries - glad you figured it out.

piccolo_admin/endpoints.py
```diff
@@ -305,6 +308,9 @@ def __init__(
         rate_limit_provider: t.Optional[RateLimitProvider] = None,
         production: bool = False,
         site_name: str = "Piccolo Admin",
+        logo_image: str = "logo.jpg",
+        logo_width: int = 40,
+        logo_height: int = 30,
```
dantownsend, line 313: I'm not sure if we need to specify height and width - we can just style the image so it looks OK using CSS. For example:

```css
img {
    height: 1em;
    width: auto;
}
```

dantownsend:
It looks good - thanks.

I'm not sure about using the Python logo as the default icon. I think I'd rather use something custom.

We could continue to use the font awesome icon, but converted to a PNG:

https://fontawesome.com/v5/icons/tools?style=light&s=solid

Alternatively, we could use the icon from this:

image

I've still got the original vector file somewhere - I'll export the icon as a PNG.

sinisaos:
@dantownsend Thanks. I agree with you, I just took the Piccolo starter asgi template icon. It would be best if you decide and put the icon and logo you want because it is your project after all.
Just one more question. Do we need to document this or not?

piccolo_admin/endpoints.py
sinisaos: I agree with you, but I was wondering if someone want a larger logo that we would allow them to do so. If you want, I can remove that option.

dantownsend:
@sinisaos For documentation, we can just expand upon the docstring where you put `:param logo_image:`.

Something like:

```python
:param logo_image:
    The path to an image on the server which will be used as the logo in the UI. It not specified, a default logo is used.
```

piccolo_admin/endpoints.py
```diff
@@ -585,6 +594,11 @@ def __init__(
             app=StaticFiles(directory=os.path.join(ASSET_PATH, "js")),
         )
 
+        self.mount(
+            path="/icons",
+            app=StaticFiles(directory=os.path.join(ASSET_PATH, "icons")),
+        )
```
dantownsend, line 600: I think this needs to be changed  - we should let the user have their custom image in any folder. If it has to be the `icons` folder it's problematic, as the user has to be put the file within the `piccolo_admin` folder - so every time they install a new version it gets overwritten.

piccolo_admin/endpoints.py
dantownsend: It's a nice idea, but there's the risk of it breaking the layout if the image is too large. I think we just style it with CSS to make it look nice, without any configuration.

sinisaos: Ok. I will remove it

piccolo_admin/endpoints.py
sinisaos: You're right. Can you change that and add a favicon and logo (whatever you think is best) because I have to go now.

sinisaos: I leave this path for the default logo and favicon. User can set their custom logo and favicon server path in `create_admin` with `logo_path` and `favicon_path` arguments.

dantownsend: Thanks. I'll try and merge this in tomorrow.

dantownsend: Sorry, didn't get a chance to do this yet. I caught Covid recently, and my brain is running a bit slow.

sinisaos: No worries.  I hope you get better.

sumitsharansatsangi: @dantownsend , all checks have passed. Whenever you feel better, you can merge this.

dantownsend: I'm about half way through reviewing it - should be done soon.

sinisaos:
Code from this PR is in #203

PR #203, opened 2022-08-25
Custom logo experiments

dantownsend:
Based on https://github.com/piccolo-orm/piccolo_admin/pull/202

Things I want to try out:

- [ ] Do we need a favicon? Can we just use the JPEG / PNG etc directly using `<link rel="icon">`?
- [ ] Because of the way the app is structured, we fetch the logo image almost every time the user navigates between pages. It's because `NavBar.vue` isn't part of `App.vue`. Maybe we can restructure it somehow, or just tell Vue not to re-render the `NavBar` component.
- [ ] Add the final logo (it needs to be a smaller file)


sumitsharansatsangi:
@dantownsend , @sinisaos , Was the experiment successful? If not, then at least provide an option to change title of the page.

sinisaos:
@sumitsharansatsangi I made a PR and that code works, but we have to wait for @dantownsend to change what he thinks needs to be changed.

sumitsharansatsangi:
Thanks @sinisaos for creating this PR. I'll use it. @dantownsend  is probably thinking of performance and is much appreciable.
His change will be awaited.

dantownsend:
Closing for now. Might come back to it at some point.

Having the custom logo was quite a bit of trouble - mostly because of the component hierarchy in the app, the nav bar was re-rendered quite frequently, and it was pull the logo from the server.

Issue #207, opened 2022-08-29
Fix bug with custom forms and media columns


dantownsend:
Custom forms don't work because `media_columns` is undefined and the form input fields don't render (`Error in render: "TypeError: Cannot read properties of undefined (reading 'media_columns')`).

![custom_forms](https://user-images.githubusercontent.com/30960668/187284317-ec8aec1f-5ae7-4c74-bede-9b61f8fa5d7c.png)

_Originally posted by @sinisaos in https://github.com/piccolo-orm/piccolo_admin/issues/85#issuecomment-1230783463_

PR #208, opened 2022-08-29
fix custom forms

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_admin/issues/207


dantownsend:
@sinisaos I think this is fixed now. The problem was `InputField` was reading `schema` from the store - the schema is for the current table, not the custom form.

Issue #85, opened 2021-08-13
Automate running of Cypress tests


dantownsend:
Now we have some basic Cypress tests, it is worth investigating how we can run these as part of GitHub actions.

Cypress tests can be notoriously slow, so we might not want to run them on every commit - we will have to see what the performance is like.

LeMeteore:
What about something like:
```
on:
    release:
        types:
            - created

jobs:
    cypress-tests:
        name: "Run Cyress tests"
        runs-on: "ubuntu-latest"
        steps:
              run:   # command to run cypress
```

Or something like:
```
on: [push]
jobs:
  cypress-tests:
    name: "Run Cyress tests"
    if: github.actor == 'foobar'
    runs-on: ubuntu-latest
    steps:
      - run: # command to run cypress tests
```

[Doc](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution)

dantownsend:
@LeMeteore Yes, that could work. We need to integrate it with this somehow: https://github.com/cypress-io/github-action

What makes this a bit more challenging is in order to run the Cypress tests, we also need the Python REST API running. There's some info on that [here](https://piccolo-admin.readthedocs.io/en/latest/contributing/index.html). cc @sinisaos who originally wrote the Cypress tests.

LeMeteore:
This if what I've tried so far. 
- [the yaml file](https://github.com/LeMeteore/piccolo_admin/blob/github-action-for-cypress/.github/workflows/cypress.yaml)
- [the error](https://github.com/LeMeteore/piccolo_admin/runs/8074848398?check_suite_focus=true):

```
start server "python -m piccolo_admin.piccolo_admin.example --inflate=10000 & command "python -m piccolo_admin.piccolo_admin.example --inflate=10000 &"
current working directory "/home/runner/work/piccolo_admin/piccolo_admin/admin_ui"
/opt/hostedtoolcache/Python/3.7.13/x64/bin/python -m piccolo_admin.piccolo_admin.example --inflate=10000 &
/opt/hostedtoolcache/Python/3.7.13/x64/bin/python: Error while finding module specification for 'piccolo_admin.piccolo_admin.example' (ModuleNotFoundError: No module named 'piccolo_admin')
/home/runner/work/_actions/cypress-io/github-action/v4/dist/index.js:3582
                error = new Error(`The process '${this.toolPath}' failed with exit code ${this.processExitCode}`);
```
[...]

PR #209, opened 2022-08-30
update Cypress tests

sinisaos:
Related to #85

### DIFF
--- a/admin_ui/cypress/integration/LoginAndLogout.spec.ts
+++ b/admin_ui/cypress/integration/LoginAndLogout.spec.ts
@@ -7,6 +7,12 @@ context('Startup', () => {
 
     it('Fill login form and redirect to admin page and then logout', () => {
 
+        cy.on("uncaught:exception", (err, runnable) => {
+            // returning false here prevents Cypress from
+            // failing the test
+            return false
+        })
+
         // Fill the username
         cy.get('[name="username"]')
             .type('piccolo')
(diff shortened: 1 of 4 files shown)


dantownsend:
@sinisaos Thanks for this.

I get something weird when running them locally. If I run the tests one at a time, then they pass. But if I run them all at once, then one of them fails:

Screenshot 2022-08-30 at 06 59 22

sinisaos:
@dantownsend Try it now. When we go back to the login page after logging out, we get the exception `NavigationDuplicated: Avoided redundant navigation to current location: "/login?nextURL=%2F"` and it blocks the tests when they all run at once.

admin_ui/cypress/integration/LoginAndLogout.spec.ts
```diff
@@ -42,6 +48,8 @@ context('Startup', () => {
         cy.location('pathname', { timeout: 2000 }).
             should('eq', '/');
 
+        cy.wait(3000);
```
dantownsend, line 51: I think we can drop this, as it's at the end of the test shouldn't be required.

sinisaos: I think we should leave this line because without this line I get an error in the test. Do you want to remove the delete tests or is restarting the server ok?

dantownsend: OK, fair enough - let's keep it for now then.

PR #214, opened 2022-09-02
Add Github action to run Cypress tests. Refs #85

LeMeteore:


sinisaos:
@dantownsend The tests run locally just fine. I don't know what the problem could be. `cy.visit` has a default timeout of 60 seconds and I think that should be enough for the page to load. Maybe we can try to remove `--inflate` form start command, maybe that can help.

dantownsend:
@sinisaos Good idea - we don't really need inflate for the current Cypress tests.

LeMeteore:
I am watching the screenshots and I'm seeing an "Apache2 Ubuntu default page" :\ It sounds like a server is running and serving piccolo_admin, but not the server we think about :smile:

LeMeteore:
I am watching the screenshots and I'm seeing an "Apache2 Ubuntu default page" :\ It sounds like a server is running and serving piccolo_admin, but not the server we think about and not the content we think about.

dantownsend:
That's interesting - I'm surprised that GitHub Actions have Apache running.

LeMeteore:
> That's interesting - I'm surprised that GitHub Actions have Apache running.

Indeed that's interesting. Either Apache is still running, either Nginx is just serving old content and not really serving the configured virtualhost. :\

LeMeteore:
I have found the problem (and I should have thougth about it :angry: ) There was a virtualhost configured as the default server and this vhost was serving all the old Apache2 content :smile: I suppressed the default default vhost, set ours as the default vhost and I'm progressing.... to a 502 Bad Gateway :smile: Somebody to reread what I've written in the vhost config ? Did I miss something
Startup -- Fill custom forms -- before each hook (failed)

dantownsend:
502 usually means that whatever is running behind the proxy (Nginx in this case) isn't responding. Maybe the Vue app isn't running for some reason?

dantownsend:
I've been playing around with this in a separate branch. Finally managed to get the tests passing: https://github.com/piccolo-orm/piccolo_admin/pull/215

I couldn't work out how to stop Apache, so used Vue's dev server as a proxy instead.

There's a lot of noise in the output:

https://github.com/piccolo-orm/piccolo_admin/runs/8177955044?check_suite_focus=true

It's because Vue's dev server uses Webpack for module building, and it spams a lot information.

I have one other idea I'd like to try. Rather than running the Vue dev server, we could just compile the front end assets (using `npm run build`), and then just run the Python server.

dantownsend:
I've merged this in now, via a fork of this PR. 🥳 

@LeMeteore Thanks so much for this. It's something which needing doing for a long time, and gives us much more confidence when making future changes.

@sinisaos Thanks a lot of your help too. Just be aware that the Nginx reverse proxy isn't required any more. I've updated the contributing docs to reflect this.

I'll see if I can add more Cypress tests over the coming weeks!

LeMeteore:
> I couldn't work out how to stop Apache, so used Vue's dev server as a proxy instead. 

Normally, Apache was not running. Nginx was running and serving old Apache content.

I'm reading your comment [here](https://github.com/piccolo-orm/piccolo_admin/pull/215#issuecomment-1236325464) and I'm realizing that I was not using the right porst in my virtualhost configuration :crying_cat_face: instead of 8080 I've written 80 

> Thanks so much for this. It's something which needing doing for a long time, and gives us much more confidence when making future changes. 

I'm Glad \o/

dantownsend:
> Normally, Apache was not running. Nginx was running and serving old Apache content.

I didn't think of that - makes sense now!

PR #215, opened 2022-09-04
Cypress GitHub actions testing

dantownsend:
Based off https://github.com/piccolo-orm/piccolo_admin/pull/214

Just seeing if I can work out the 502 error.


dantownsend:
In the tests we have:

 * The backend server running on port `localhost:8000` (this is the default host and port of Hypercorn, which is used to serve the app in [example.py](https://github.com/piccolo-orm/piccolo_admin/blob/master/piccolo_admin/example.py))
 * The Vue app running on port `piccolo_admin:8080` ([vue.config.js](https://github.com/piccolo-orm/piccolo_admin/blob/master/admin_ui/vue.config.js))
 * Nginx running with a virtual host called `piccolo_admin`

Cypress `baseUrl` is currently set to `localhost:3000` - should this be `piccolo_admin` instead?

dantownsend:
It's working now - we build the front end assets, and then just run the Python server to test against.

Issue #213, opened 2022-09-02
Improve error messages in Piccolo Admin


dantownsend:
If we get an error response back from the server when saving a row, we show the response in the UI.

This is if we get a validation error from Pydantic:

<img width="702" alt="Screenshot 2022-09-02 at 12 13 43" src="https://user-images.githubusercontent.com/350976/188127742-8d808366-d2b0-4b03-bc4b-739289e5e2e9.png">

This is if we get a database error:

<img width="748" alt="Screenshot 2022-09-02 at 12 14 03" src="https://user-images.githubusercontent.com/350976/188127784-b483b5a7-7095-43dc-b466-71aec736615f.png">

It would be good if we parsed these errors where possible, and created an improved UI.

Related to https://github.com/piccolo-orm/piccolo_admin/issues/167

sinisaos:
@dantownsend Maybe the error messages could look like this?

![db_errors](https://user-images.githubusercontent.com/30960668/188227966-20517961-5bdf-4819-9fec-db14495be77c.png)

and pydantic validation errors like this

![pydantic](https://user-images.githubusercontent.com/30960668/188228571-a4e32ff4-e194-4a58-8482-e6ca292af8c1.png)

dantownsend:
Yes, something like that. I think the only problem is it looks visually very similar to a button.

Maybe red text will work, without a background?

sinisaos:
If we use only the text, then we can display the entire error message so the user knows what the problem is, like this

![pydantic_error](https://user-images.githubusercontent.com/30960668/188259124-2ca378a3-b3a2-4527-9537-e510f124ca2a.png)

or 

![db_error](https://user-images.githubusercontent.com/30960668/188259135-4d1dd376-8736-42db-9fab-285ae1b753de.png)

sinisaos:
Or maybe like this so it looks different from the button.

![error](https://user-images.githubusercontent.com/30960668/188699721-f6d58c0e-514a-4cff-b2ba-479f9827d558.png)

dantownsend:
Yeah, that could work. Do you think we need the close button? It balances the design nicely, but I wonder if someone would want to hide it.

How about adding a warning icon too?

![188699721-f6d58c0e-514a-4cff-b2ba-479f9827d558](https://user-images.githubusercontent.com/350976/188729609-33739dc2-027e-4a98-8378-d21eaeda8c4e.png)

sinisaos:
[...]

PR #219, opened 2022-09-07
add error messages

sinisaos:
Related #213


admin_ui/src/components/AddRowForm.vue
```diff
@@ -66,7 +68,13 @@ export default {
                     data: json
                 })
             } catch (error) {
-                const data = error.response.data
+                let data = error.response.data
+
+                if (data["db_error"]) {
+                    data = data["db_error"]
+                } else {
+                    data = data["detail"][0]["msg"]
```
dantownsend, line 76: If there are multiple errors, does this mean we're just showing the first error?

How about we pass in a list of errors to the `RowFormErrors` component instead.

admin_ui/src/components/EditRowForm.vue
```diff
@@ -110,7 +112,13 @@ export default Vue.extend({
...
                 this.$store.commit("updateApiResponseMessage", message)
             } catch (error) {
-                const data = error.response.data
+                let data = error.response.data
+
+                if (data["db_error"]) {
+                    data = data["db_error"]
+                } else {
+                    data = data["detail"][0]["msg"]
+                }
```
dantownsend, line 121: Lower down in the file we have `if (typeof data != "string") {`

We should try and merge both pieces of logic.

dantownsend: Screenshot 2022-09-07 at 11 16 59

If the error message is from Pydantic, we also need to use the `loc` value so we can tell the user which field has the problem.

admin_ui/src/components/RowFormErrors.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+    <div class="errors">
+        <font-awesome-icon icon="exclamation-triangle" />
+        <span>
+            <b> {{ errors }}</b>
+        </span>
+    </div>
```
dantownsend, line 7: We might have to change this so it's a list.

Something like:

```vue
<p><font-awesome-icon icon="exclamation-triangle" />Error</p>
<ul>
    <li v-for="error in errors" :key="error">{{ error }}</li>
</ul>
```

admin_ui/src/components/EditRowForm.vue
dantownsend: I'd consider adding a utility function like `parsePydanticError` or something, so we can re-use the logic for converting the error into a nice string.

dantownsend: We might also be able to distinguish between Pydantic, database errors, and other errors based on the status code.

admin_ui/src/components/AddRowForm.vue
sinisaos: Yes. Now we have that situation. We only show one error message at a time, but that makes sense because at that point we only have one status code (422 for db errors or 500 for Pydantic errors) from the server response. Sorry if I am wrong.

admin_ui/src/components/EditRowForm.vue
sinisaos: As I understand it, we have two types of messages (database error from Piccolo API `db_exception_handler` with status code 422 and Pydantic validition errors). I can add `loc` to the Pydantic error message

admin_ui/src/components/AddRowForm.vue
dantownsend: You're right - there's only one response from the server. But lets say there's a table like this:

```python
class MyTable(Table):
    json_1 = JSONB()
    json_2 = JSONB()
```

And you tried posting back `{'json_1': 123, 'json_2': 456}`. The error message that Pydantic returns will show an error for `json_1` and `json_2`.

admin_ui/src/components/RowFormErrors.vue
sinisaos: It's a great idea, but as I wrote in previous comments, we only get one error from the server if I understand correctly.

admin_ui/src/components/AddRowForm.vue
dantownsend: Here's an example - I triggered two errors in the form (one because the name field is too long, and the other because the JSON is incorrect):

Screenshot 2022-09-07 at 15 23 05

sinisaos: You are right. I have to try that, but I can't right now. I'll be able to later tonight.

dantownsend: No worries

sinisaos:
@dantownsend I have made the changes you suggested. I hope it's better now.

validation_error

db_error

admin_ui/src/components/AddRowForm.vue
```diff
@@ -67,18 +69,26 @@ export default {
...
+                const validationError = data["detail"]
+                let errorsArray = []
+
+                if (databaseError) {
+                    errorsArray.push(databaseError)
+                } else {
+                    validationError.forEach((item: any) => {
+                        errorsArray.push(`Field ${item.loc[1]} - ${item.msg}`)
+                    })
+                }
```
dantownsend, line 82: We need to be careful here, as we can get other types of error back.

For example, if it's a 500 error we just get a string back saying '500 error'.

Screenshot 2022-09-08 at 15 24 51

Screenshot 2022-09-08 at 15 24 35

I think it should check status codes.

 * Pydantic validation errors return a 422.
 * Database unique errors are also returned as 422.

If the status code is something else, we should just render what ever the response was.

```javascript
const data = error.response.data

let errorsArray: string[] = []

if (error.response.status == 422) {
    const databaseError = data["db_error"]
    const validationError = data["detail"]

    if (databaseError) {
        errorsArray.push(databaseError)
    } else {
        validationError.forEach((item: any) => {
            errorsArray.push(`Field ${item.loc[1]} - ${item.msg}`)
        })
    }
} else {
    errorsArray.push(data)
}
```

admin_ui/src/components/EditRowForm.vue
```diff
@@ -111,18 +113,26 @@ export default Vue.extend({
...
+                const validationError = data["detail"]
+                let errorsArray = []
+
+                if (databaseError) {
+                    errorsArray.push(databaseError)
+                } else {
+                    validationError.forEach((item: any) => {
+                        errorsArray.push(`Field ${item.loc[1]} - ${item.msg}`)
+                    })
+                }
```
dantownsend, line 126: As we're repeating the logic from above, we should try and move it to shared function called something like `parseErrorResponse`.

dantownsend:
That looks better, thanks. Left a couple of comments.

admin_ui/src/components/AddRowForm.vue
dantownsend: We can also get a 405 status code when running in read only mode.

Screenshot 2022-09-08 at 15 52 02

dantownsend:
Thanks. I made a couple of changes - I used the new error component on the custom forms pages too. And tweaked the styles for light mode.
Screenshot 2022-09-08 at 23 34 29
Screenshot 2022-09-08 at 23 34 22

Issue #221, opened 2022-09-08
Write Cypress test for bulk updates


dantownsend:
### Why

We're trying to increase the test coverage we have using Cypress - it helps us develop with much more confidence.

### What the test needs to do

Tick some checkboxes:

<img width="1063" alt="Screenshot 2022-09-08 at 13 08 55" src="https://user-images.githubusercontent.com/350976/189118488-069c4021-4ba0-4725-9ced-d2fceee30f88.png">

Click on the `update 8 rows` button:

<img width="1059" alt="Screenshot 2022-09-08 at 13 09 21" src="https://user-images.githubusercontent.com/350976/189118485-7bba9c8f-3fe0-4dc7-9f96-90c328662ae3.png">

Then submit the form:

<img width="1058" alt="Screenshot 2022-09-08 at 13 09 50" src="https://user-images.githubusercontent.com/350976/189118478-980e390e-b2ee-4053-85b4-6d7c828697cd.png">

And make sure the values in the table were updated.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_admin/pull/230

PR #230, opened 2022-09-08
bulk update Cypress test

sinisaos:
Related to #221


admin_ui/cypress/e2e/BulkUpdate.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        // Choose from selects
+        cy.get('select[name=property]')
+            .select('gender');
+
+        cy.get('select[name=gender]')
+            .select('f');
+
+        // Locate and submit the form
+        cy.get('form').submit();
```
dantownsend, line 50: This looks great, thanks!

I wonder if it's possible to add an extra step which checks whether the values have updated?

sinisaos: I pushed another commit. I hope you meant something like that.

dantownsend: That's it, thanks!

Issue #216, opened 2022-09-04
Write a Cypress test for media uploads


dantownsend:
We recently added media upload support to Piccolo Admin.

You can see here in the UI (running `python -m piccolo_admin.example`):

<img width="698" alt="Screenshot 2022-09-04 at 23 44 35" src="https://user-images.githubusercontent.com/350976/188336270-bed73f51-4dcf-4975-8d45-a8a109104e8d.png">

It would be good to have Cypress tests for these. To learn more, see our [contributing docs](https://piccolo-admin.readthedocs.io/en/latest/contributing/index.html).

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_admin/pull/231

PR #231, opened 2022-09-10
Cypress file upload test

sinisaos:
Related to #216


dantownsend:
Thanks for this. I wasn't sure how image uploading would work in Cypress, but seems to work well.

I made the file slightly smaller, and increased the timeout, just to give the file the best chance possible to finish uploading. On my local machine I have `example.py` configuered to upload files to S3, so is a bit slower.

dantownsend:
Do you think it's worth adding a test for array columns which are using media storage?

sinisaos:
Great, thanks I always forget to use smaller files. I tried it this morning and it works without problems thanks to the [selectFile](https://www.cypress.io/blog/2022/01/19/uploading-files-with-selectfile/) command. Previously, a plugin (cypress-file-upload plugin) should have been used.

sinisaos:
> Do you think it's worth adding a test for array columns which are using media storage?

The file upload is the same for uploading single and multiple files (for Array column we also upload one file at a time), so I don't think there is a need for it.

dantownsend:
OK, fair enough - lets merge this in.

Issue #226, opened 2022-09-08
Cypress - restart Python server before each test


dantownsend:
Our process for running Cypress tests at the moment is:

```
# Launch the Python server
python -m piccolo_admin.example

# Run the Vue App
cd admin_ui
npm run serve

# Run Cypress
npm run cypress:run
```

Ideally, between each test we would restart the Python web server, so we start from a blank slate each time. Otherwise some tests may modify the database which effects later tests.

- [ ] Work out if it's possible
- [ ] If so, implement it

sinisaos:
@dantownsend I played around with this a bit but couldn't find a way to do it. I tried `cy.exec()` (though it's considered an anti-pattern) but it didn't work. In the end, I don't think it matters that much, because when we do tests locally, I don't think it's a problem to start the server when the tests are done, and when the tests are done in GH actions with each commit server is restarted automatically (if I understand correctly) .

dantownsend:
Thanks for giving it a go.

One way it's sometimes done is having a REST API endpoint, and Cypress calls that. So `example.py` might have a `/reset` endpoint, which resets the database.

I've been playing around with Playwright. It's an alternative to Cypress, and tests can be written in Python, which would make things like this easier, as we can directly call other Python code from within our tests:

https://playwright.dev/

We could run a test, and then check the results are correct in the database.

I'm not saying we should switch right now, but it's an option if we run into difficult use cases.

sinisaos:
Playwright looks great and seems to have all the features as Cypress but code can be written in Python which is great.

> One way it's sometimes done is having a REST API endpoint, and Cypress calls that. So `example.py` might have a `/reset` endpoint, which resets the database.

I like that idea although I don't know how to restart server from endpoint.

dantownsend:
I don't think we could restart the server, but the `reset` endpoint would delete the data from the database:

```python
await Movie.delete(force=True)
...
```

And call this again:

https://github.com/piccolo-orm/piccolo_admin/blob/13693161f237025a4f715abd4b4bbeda3bda026b/piccolo_admin/example.py#L350

If I had a time machine, I would probably go back and use Playwright instead of Cypress. I still really like Cypress, but being able to use Python directly in our tests gives us way more options.

sinisaos:
Playwright looks interesting. I'll try to rewrite the Cypress tests (we don't have a lot of tests) and see how it works.

dantownsend:
Yeah, have a play around and see what you think.

[...]

PR #234, opened 2022-09-22
Playwright tests

sinisaos:
Related to #226. `playwright install` needs to be run to install the required browsers. The tests are used by running `./scripts/dev-server` in one terminal and in second terminal run tests with `./scripts/run-e2e-test.sh`.


e2e/test_playwright.py
```diff
@@ -0,0 +1,337 @@
...
+    Director.insert(*[Director(**d) for d in DIRECTORS]).run_sync()
+
+
+def test_login_logout(playwright: Playwright, reset_database) -> None:
+    browser = playwright.chromium.launch()
+    context = browser.new_context()
+    # Open new page
+    page = context.new_page()
+    # Go to http://localhost:3000/#/login?nextURL=%2F
+    page.goto("http://localhost:3000/#/login?nextURL=%2F")
```
dantownsend, line 19: Out of interest, did you generate these using Playwright Codegen?

If so, how did you find it?

scripts/run-tests.sh
```diff
@@ -9,4 +9,4 @@ mkdir -p piccolo_admin/dist/css
 mkdir -p piccolo_admin/dist/js
 touch piccolo_admin/dist/index.html
 
-python -m pytest --cov=piccolo_admin --cov-report xml --cov-report html --cov-fail-under 85 -s $@
+python -m pytest tests --cov=piccolo_admin --cov-report xml --cov-report html --cov-fail-under 85 -s $@
```
dantownsend, line 12: Hmm, this is a tricky one.

 I guess you added this so Pytest wouldn't also run the Playwright tests.

The only downside is we can no longer specify a specific test to run (passed in via $@).

```
./run-tests.sh tests/test_foo.py
```

dantownsend:
Things to do:

- [ ] We need to update `contributing.rst` so it mentions Playwright rather than Cypress
- [ ] Remove Cypress
- [ ] Add GitHub Actions

e2e/test_playwright.py
sinisaos: Sorry for the late reply. Yes, tests are generated from `playwright codegen localhost:3000`. After that, two windows were opened. One where you can interact with the web page and one where the code is generated. I had to change each test a bit later, but the tests are 80% generated with Playwright and I think it's a great feature that is very useful.

scripts/run-tests.sh
sinisaos: Yes, you're right. I had to add that because `pytest` runs both tests at once.

sinisaos:
@dantownsend Setting up Playwright in GH actions was not an easy task. The main problem was running the dev server and the tests in parallel at once. Luckily I found this [snippet from Simon Willison](https://til.simonwillison.net/pytest/playwright-pytest) which I modified a bit and added a database reset. This allows us to start the dev server and tests in parallel with `./scripts/run-e2e-test.sh`. The second step was to set up GH actions. For this I found a great tool to run gh actions locally called [act](https://github.com/nektos/act) and it is very useful.

I have one more question. Do we even use Storybook or can I remove it as well to reduce npm dependencies?

e2e/test_playwright.py
```diff
@@ -0,0 +1,370 @@
...
+            response = conn.getresponse()
+            if response is not None:
+                yield process
+                break
+        except ConnectionRefusedError:
+            time.sleep(1)
+            retries -= 1
+
+    Director.delete(force=True).run_sync()
+    Director.insert(*[Director(**d) for d in DIRECTORS]).run_sync()
```
dantownsend, line 36: We might not need to do this anymore - it looks like each test starts the server again, so it will already have fresh data each time.

scripts/run-e2e-test.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Run end-to-end tests
+
+export PICCOLO_CONF="e2e.piccolo_conf"
+
+python -m pytest e2e
```
dantownsend, line 6: One of the main differences between Playwright and Cypress is Playwright doesn't emit images / videos by default. We need to pass in the `--screenshot=on`, and `--video=on` flags.

https://playwright.dev/python/docs/test-runners#cli-arguments

It would be nice to upload these as build artefacts (not sure if screenshots/videos/both, and whether just on failure). It will help us see why tests fail.

The Cypress GitHub action had some code we can reuse:

```
      - name: "Upload Cypress Screenshots and Videos"
        uses: actions/upload-artifact@v3
        with:
          name: cypress-artifacts
          path: |
            /home/runner/work/piccolo_admin/piccolo_admin/admin_ui/cypress/videos/
            /home/runner/work/piccolo_admin/piccolo_admin/admin_ui/cypress/screenshots/
        if: always()
```

e2e/test_playwright.py
dantownsend: Let's keep Storybook for now - otherwise this PR gets quite big.

Maybe we drop it in the future. I'm torn on Storybook. It can be useful during development sometimes, and encourages us to write nicely isolated components. But it's a pain to configure using Webpack.

scripts/run-e2e-test.sh
sinisaos: Ok, I'll try adding that and see how it works. Now [act](https://github.com/nektos/act) comes very handy because I can try everything locally, not by pushing commits to Github.

sinisaos:
> I did a little bit of work on Storybook fairly recently because it was broken. Honestly, I barely use it, and getting webpack to work properly with it is a pain. I'm OK with keeping it for now, but might drop it in the future.

OK, I won't delete Storybook.

scripts/run-tests.sh
dantownsend: Maybe instead of specifying 'tests', we can ignore the integration folder instead.

Something like `pytest --ignore=e2e`

scripts/run-e2e-test.sh
sinisaos: I tried `--screenshot=only-on-failure` and `--screenshot=on` . I made `screenshots` folder inside `e2e` and use
```yaml
- name: "Upload Playwright Screenshots"
  uses: actions/upload-artifact@v3
  with:
    name: playwright-artifacts
    path: |
      /home/runner/work/piccolo_admin/piccolo_admin/e2e/screenshots/
  if: always()
```
and 
```yaml
- uses: actions/upload-artifact@v2
  if: always()
  with:
    name: screenshots
    path: screenshots/
    retention-days: 5
```
but result is always same. I get this warning `::warning::No files were found with the provided path: /home/runner/work/piccolo_admin/piccolo_admin/e2e/screenshots/. No artifacts will be uploaded.` and `screenshots` folder is empty.
In the end I put `--screenshot=only-on-failure`. Do you have any idea how to save screenshots?

dantownsend: I think the folder name might be `test-results' instead of `screenshots`.

For some other Playwright tests I have, the path looks like this:

```
test-results/integration-test-signup-py-test-signup-chromium/test-failed-1.png
```

I think it' possible to customise the path using the `--output` flag.

sinisaos: Sorry but I was unable to set the `test-results`. I don't know where the problem is and I always get `::warning::No files were found with the provided path` on `act` tool. I tried to put `test-results` in `e2e` and in `root`. I tried `--output` flag but nothing works. GH actions says that artifact is successfully uploaded but files size is 0. If you can fix it somehow, that would be great. Apart from that (all tests and GH actions) everything else works fine.

sinisaos:
@dantownsend I ended up using videos and the video artifacts were uploaded successfully.

dantownsend:
@sinisaos That's great - videos are better anyway.

sinisaos:
@dantownsend If you have nothing to add to this PR feel free to merge this.

dantownsend:
Cool, it looks good to me.

PR #244, opened 2022-11-05
Adding support for custom callbacks per table

haffi96:
- The general idea to allow custom logic being run for each table, this is more of a POC. So if there is anything wrong with doing this. Let me know.
- First time working with vue and typescript, so would be great to have some suggestions for improvement

For this to work, would need this https://github.com/piccolo-orm/piccolo_api/pull/200

image

### DIFF
--- a/admin_ui/src/components/CallbackButton.vue
+++ b/admin_ui/src/components/CallbackButton.vue
@@ -1,0 +1,55 @@
+<template>
+    <a
+    href="#"
+    id="callback"
+    title="Row callback"
+    v-on:click.prevent="$emit('triggered')"
+    >
+        <span v-if="includeTitle">{{ $t("Callback") }}</span>
+</a>
+</template>
+
+
+<script lang="ts">
+export default {
+    props: {
+        includeTitle: {
+            type: Boolean,
+            default: false
+        }
+    }
+}
+</script>
+
+<style scoped lang="less">
+@import "../vars.less";
+
+@keyframes shake {
+    0% {
+        transform: rotate(0deg);
+    }
+    50% {
+        transform: rotate(-10deg);
+    }
+    100% {
+        transform: rotate(10deg);
+    }
+}
+
+a#callback {
+    text-decoration: none;
+
+    span {
+        padding-left: 0.2rem;
+    }
+
+    &:hover {
+        color: @red;
+
+        svg {
+            animation: shake 0.4s;
+            animation-iteration-count: infinite;
+        }
+    }
+}
+</style>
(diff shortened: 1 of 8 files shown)


dantownsend:
@haffi96 This is cool, thanks.

I agree that it's something we need to add.

Some initial thoughts:

* Rather than callbacks, how about we call them 'actions'? I might be biased on the name, as this is what I used in the past with Django.
* It would be good if multiple actions / callbacks could be registered for a single table.  So the the API call be be `POST /actions/<some-action-name/`.
* At the moment it seems like the action can only accept a single row as an argument, but it would be great if multiple rows could be passed in.

haffi96:
> Rather than callbacks, how about we call them 'actions'?
haha actions was actually my initial choice too (inspired by flask-admin for myself 🤓 ). Decided against it as i saw that there was already some `actions` related components so didnt want to overlap. But i agree, actions sounds better. Will update

> It would be good if multiple actions / callbacks could be registered for a single table
👍🏽 

> At the moment it seems like the action can only accept a single row as an argument, but it would be great if multiple rows could be passed in.

Yep, that was intentional as i wasn't sure if waiting for the response for an action to process for multiple rows may be kinda weird. But will experiment with this. I agree, would definitely be nice to have it for multiple rows

haffi96:
Can register multiple actions per table now and also select multiple rows at once. 

Atm, the admin only sends this -> `{"table_name": ..... , "row_ids": [...]}` back to the user registered custom action functions

And as you can see i've struggled abit with the css 😆 Will have another go at better aligning the drop down buttons

image

dantownsend:
@haffi96 Cool, thanks! The UI looks good.

I think passing just the row IDs should be OK.

dantownsend:
@sinisaos What do you think about this? I think it could be a really useful feature.

dantownsend:
Still valid.

haffi96:
Hey @dantownsend sorry about the delay. I've done some more work on this last and implemented your suggestions on the piccolo api PR. Just have some unit tests left to do. 

But quick question, Is this PR dependent on https://github.com/piccolo-orm/piccolo_api/pull/145 being merged first?

sinisaos:
> But quick question, Is this PR dependent on [piccolo-orm/piccolo_api#145](https://github.com/piccolo-orm/piccolo_api/pull/145) being merged first?

@haffi96 No, this PR is not dependent  [piccolo-orm/piccolo_api#145](https://github.com/piccolo-orm/piccolo_api/pull/145) . It's a completely different PR that provides a solution for propper bulk actions (delete and update) for Piccolo API and Piccolo Admin.

Issue #254, opened 2022-11-15
Fix tests


dantownsend:
A lot of the test suite has blown up with the latest version of Starlette, which uses `httpx` instead of `requests` for its test client.

We will have to fix the tests, and make `starlette=0.21.0` the minimum from now on, because supporting test suites for the older versions will be a massive pain.

sinisaos:
@dantownsend If you want I can try to fix the tests. I haven't worked with the new Starlette TestClient yet, so I'd like to try it.

dantownsend:
@sinisaos Yes, feel free to give it a go.

The TestClient is mostly the same, but has subtle differences for cookies and how it handles redirects.

For example, when calling the `/login/` endpoint, you have to use `follow_redirects=False`, otherwise the response you get will be the redirected one, and it won't contain the session auth cookie. I hope that makes sense.

Also, the `Response.cookies` object is slightly different, which can cause some issues.

sinisaos:
@dantownsend OK. I will try and use the Piccolo API tests as a reference. If I didn't succeed, I will reported here.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_admin/pull/255

PR #255, opened 2022-11-15
fix test for Starlette new test client

sinisaos:
Related to #254


sinisaos:
@dantownsend Fortunately, there was only one small change in tests. Nothing like in Piccolo API.

dantownsend:
@sinisaos Phew, that's a relief!

Thanks for this.

Issue #257, opened 2022-11-16
Support Python 3.11


dantownsend:
Update setup.py and tests.

PR #258, opened 2022-11-16
Add Python 3.11 support

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_admin/issues/257

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -42,7 +42,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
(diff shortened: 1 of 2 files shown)


sinisaos:
@dantownsend Could you please correct this line `app = create_admin(tables=[MyTable]  debug=True)` in `https://github.com/piccolo-orm/piccolo_admin/blob/master/docs/source/debugging/index.rst`.
I think it should be `app = create_admin(tables=[MyTable], debug=True) `

dantownsend:
@sinisaos You're right - good catch. I'll do it in a separate PR, so it doesn't trigger GitHub's CodeQL, as it takes ages to finish!

PR #259, opened 2022-11-18
Fixed exclude_visible_columns for TableConfig

web-maker:
### What?

Fix for issue with `exclude_visible_columns` for `TableConfig`.

### Why?

Right now `exclude_visible_columns` parameter is not working if the primary key column is not in the list of excluded columns.  This PR introduces a possible quick fix for that problem.


dantownsend:
@web-maker Good catch!

It's quite an interesting bug. Turns out using `in` with a generator causes some weird behaviour ...

Issue #262, opened 2022-12-13
Make admin more flexible for subclassed `BaseUser`


dantownsend:
In some situations the user may need to modify `BaseUser` and `SessionsBase` to use `UUID`s instead.

We can do this with a very minor change to Piccolo Admin.

```python
class CustomSessions(SessionsBase):
    id = UUID(primary_key=True)
    user_id = UUID(null=False)

class CustomUser(BaseUser):
    id = UUID(primary_key=True)

create_admin(
    tables=[TableA, TableB],
    auth_table=CustomUser,
    session_table=CustomSessions
)
```

This pretty much works out of the box, except for one minor thing here:

https://github.com/piccolo-orm/piccolo_admin/blob/f2fbcea0ef5dcff029f67a023b1d1965e7a07fec/piccolo_admin/endpoints.py#L776-L780

We just need to cast `user_id` to a string, in case it's a `UUID`.

sinisaos:
@dantownsend You're right. The only change we need to make is to cast the `user_id` to `user_id=str(request.user.user_id)` and everything will work for both `Serial` or `UUID` primary key.

dantownsend:
@sinisaos Yeah, I think it's worth adding as it's such a small change.

PR #263, opened 2022-12-13
allow `BaseUser` subclass to be used, which uses a `UUID` primary key

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_admin/issues/262

### DIFF
--- a/piccolo_admin/endpoints.py
+++ b/piccolo_admin/endpoints.py
@@ -776,7 +776,7 @@ class AdminRouter(FastAPI):
     def get_user(self, request: Request) -> UserResponseModel:
         return UserResponseModel(
             username=request.user.display_name,
-            user_id=request.user.user_id,
+            user_id=str(request.user.user_id),
         )
 
     ###########################################################################


sinisaos:
@dantownsend If it's not a problem could you also add a small change to this same PR so that when we have an error in the edit or add form, we automatically scroll to the top of the screen so that the error is immediately visible to the user. It just needs to be added  `window.scrollTo(0, 0)` to [here](https://github.com/piccolo-orm/piccolo_admin/blob/f2fbcea0ef5dcff029f67a023b1d1965e7a07fec/admin_ui/src/components/AddRowForm.vue#L72-L85) and [here](https://github.com/piccolo-orm/piccolo_admin/blob/f2fbcea0ef5dcff029f67a023b1d1965e7a07fec/admin_ui/src/components/EditRowForm.vue#L121-L134). Thanks

dantownsend:
@sinisaos Yes, sounds like a good idea - will do it in a separate PR, and will merge it in.

PR #268, opened 2022-12-25
make the first visible column a crud link

### LINKED ISSUE #266: When the id field is hidden in TableConfig, there is no way to edit a record
So we have an UUID primary keys and want to hide them in admin, but that's happens:

image

It would be cool if the record link field was parameterized in `TableConfig`

sinisaos:
Resolves #266


sinisaos:
@dantownsend Can you merge this PR too? A small change to make the first visible column a link. That's also close #266

dantownsend:
@sinisaos Yeah, this is a tricky one.

The problem with automatically making it the first column, if that column could be a foreign key.

Do we just make it configurable in `TableConfig`? I'll have a play around.

dantownsend:
@sinisaos I made it configurable in `TableConfig` - let me know what you think.

sinisaos:
> The problem with automatically making it the first column, if that column could be a foreign key.

Yes, you're right. I didn't think about that. 

> I made it configurable in `TableConfig` - let me know what you think.

Yes, it is better, but the problem remains the same. For example, if you put `Ticket.movie` as a `link_column`  we can choose `Ticket` row, but not the foreign key of the `Movie` table.

Another solution is to make edit and delete actions more obvious (visible) something like this. 

crud_link

With that we can have the first visible column as a foreign key (and get access to fk table via link), and we can edit or delete that `Director` row via the icons. What do you think?

dantownsend:
With this new approach you can make any of the columns into the link. For example, here I made `booked_on` the link:

Screenshot 2023-01-26 at 16 13 28

So it's easier to avoid clashes with foreign key columns.

One of the main issues is `visible_columns` doesn't determine order - just whether or not a column is visible. If it did, then we wouldn't have a problem:

```python
ticket_config = TableConfig(
    table_class=Ticket,
    link_column=Ticket.booked_by,
    # I make sure that the link column is the first in the list:
    visible_columns=[Ticket.booked_by, Ticket.booked_on, Ticket.movie],
)
```

Having the edit link more prominent is a good idea. Delete is redundant anyway, as you can do that via bulk delete, and on the edit page. I don't think the ideas are mutually exclusive - we can probably do both. I just want a quick fix for this problem, as it's quiet a niche problem.

dantownsend:
I made one more change, so the order of columns in the row listing page obeys `visible_columns`.

sinisaos:
Sorry, but the problem with the foreign key as a `link_column` still remains because it cannot be a link to edit a table row and a foreign key table row at the same time. I think we should go with visible actions icons (easiest to do without changing a lot of code) and we shouldn't let the user to configure the `link_column` at all because he can accidentally put the fk column as a link and then we have a problem because that way he can't edit the fk table row what is wrong. But do as you wish.

dantownsend:
Yeah, with `link_column` we rely on the user being sensible. They could potentially break things. We could raise a `ValueError` if they set specify a `ForeignKey` as `link_column`.

sinisaos:
Good idea. With `ValueError` we can prevent the fk column from being used as a `link_column` and so the user will be warned in the terminal that he can use that. Something like this is result
```
Traceback (most recent call last):
  File "app.py", line 64, in <module>
    create_admin(
  File "/home/sinisa/test/lib/python3.8/site-packages/piccolo_admin/endpoints.py", line 1147, in create_admin
    return AdminRouter(
  File "/home/sinisa/test/lib/python3.8/site-packages/piccolo_admin/endpoints.py", line 489, in __init__
    raise ValueError(
ValueError: Cannot use foreign key column as `link column`.
```

sinisaos:
@dantownsend This looks and works great :+1:.

dantownsend:
@sinisaos Thanks - will merge it in.

PR #269, opened 2022-12-26
allow BYO auth backend

trondhindenes:
Allows replacing the built-in auth backend with a custom one.

This allowed me to do:
```python
class HackyAuthUser(BaseUser):
    def __init__(self, user_id: str = 'unknown', display_name: str = 'unknown'):
        self._user_id = user_id
        self._display_name = display_name

    @property
    def is_authenticated(self) -> bool:
        return True

    @property
    def display_name(self) -> str:
        return self._display_name

    @property
    def user_id(self) -> str:
        return self._user_id


class HackyAuthBackend(AuthenticationBackend):
    def __init__(self, header_name):
        self.header_name = header_name

    async def authenticate(self, conn):
        if self.header_name not in conn.headers:
            raise AuthenticationError('Invalid credentials')
        user_name = conn.headers[self.header_name]
        return AuthCredentials(scopes=[]), HackyAuthUser(user_name, user_name)


app = FastAPI(
    routes=[
        Mount('/admin/', create_admin(
            tables=APP_CONFIG.table_classes,
            auth_backend=HackyAuthBackend(header_name='Authorization'))
              ),
    ],
)
```
It would be cool if it was somehow possible to override the default "non-authenticated" behavior, and for example have admin-api redirect the user to another login url instead of the built-in one, but I didn't find a clean way to do that.

### DIFF
--- a/piccolo_admin/endpoints.py
+++ b/piccolo_admin/endpoints.py
@@ -43,6 +43,8 @@ from starlette.middleware.exceptions import HTTPException
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse
 from starlette.staticfiles import StaticFiles
+from starlette.authentication import AuthCredentials, AuthenticationBackend, AuthenticationError, SimpleUser
+from piccolo_api.shared.auth import UnauthenticatedUser, User
 
 from .translations.data import TRANSLATIONS
 from .translations.models import (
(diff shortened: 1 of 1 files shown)


dantownsend:
Still relevant - I want to merge this in. Just need to think about what to do with the logout endpoint.

trondhindenes:
yup, I'm not sure either. Maybe just allow pointing to a custom logout url? If this was setup using oauth proxy or similar, we'd send the user to the auth0 logout url (which is known) for the auth0 app that is used with oauth proxy. Most auth providers supply a logout url, so having that as an optional parameter would imho make sense.

PR #272, opened 2023-01-06
Fixed German Translation

hblunck:
Offering some updates to the (apparently automatically generated) German translation. I am a native German speaker.


dantownsend:
@hblunck Thanks a lot for this. You're right, these were machine translated, so definitely needed improving.

PR #278, opened 2023-01-24
table group option in the table menu

sinisaos:
Related to [this disussion](https://github.com/piccolo-orm/piccolo_admin/discussions/199). Changed Playwright tests as `page.locator` API changed in `v1.27.0`.


piccolo_admin/endpoints.py
```diff
@@ -865,11 +870,19 @@ def get_meta(self) -> MetaResponseModel:
 
     ###########################################################################
 
-    def get_table_list(self) -> t.List[str]:
+    def get_table_list(self) -> JSONResponse:
```
dantownsend, line 873: Will this give us that error we had before in FastAPI?

piccolo_admin/endpoints.py
```diff
@@ -865,11 +870,19 @@ def get_meta(self) -> MetaResponseModel:
...
         """
-        Returns a list of all tables registered with the admin.
+        Returns a list of all apps with tables registered with the admin.
         """
-        return [i.table_class._meta.tablename for i in self.table_configs]
+        app_names: t.Dict[t.Optional[str], t.List[str]] = {}
```
dantownsend, line 877: Should this be `group_names`?

piccolo_admin/endpoints.py
sinisaos: We are fine with this because that bug is fixed in [v0.89.1](https://github.com/tiangolo/fastapi/releases/tag/0.89.1).

piccolo_admin/endpoints.py
sinisaos: You're right, I'll change that.

sumitsharansatsangi:
Thanks @sinisaos  for the PR.
@dantownsend , All checks have passed. Can you please merge it, if it is good to go ?

dantownsend:
@sumitsharansatsangi I think it's almost there.

@sinisaos It's looking good. One thing I spotted is the two tables at the bottom look like they belong to the same group, but they don't - the bottom most one doesn't have a group.

Screenshot 2023-01-25 at 22 30 38

There's a couple of solutions:

1.  The ones without a group should go at the top.
2. We visually distinguish them more - for example, make the background darker if it's in a group.

sumitsharansatsangi:
@sinisaos , One thing I want to suggest is to make the group collapsible, like `Tables` header is.  
Thank you.

dantownsend:
I had a play around with it.

I made it so any tables without a group go at the top.

It's all sorted alphabetically now - first by group name, then table name.

@sinisaos What you did was really good. The changes I made were basically just sorting the sidebar. I styled the group name slightly differently to make it easier to differentiate. What do you think?

Screenshot 2023-01-26 at 11 39 29

dantownsend:
I've made the groups collapsible. If you're both happy with this, I'll release it and we can always improve upon it in future PRs.

sinisaos (commented):
@dantownsend This is awesome, looks much better than before.

dantownsend:
@sinisaos Thanks, will merge it in.

sumitsharansatsangi:
@dantownsend , please release it.

PR #283, opened 2023-01-28
add `order_by` to `TableConfig`

sinisaos:
Related to #280 

Remaining tasks:

- [x] Add more tests for Playwright (sorting by multiple columns), and that specifying a sorting saves it in the URL.
- [x] Set minimum `piccolo_api` version in `requirements.txt` (should be the version which supports multiple order by args).
- [x] Indicate on the table which columns are being sorted.


sumitsharansatsangi:
Thanks @sinisaos , for the PR.
@dantownsend , This PR has passed all the test, Can you please merge it?

sinisaos:
@sumitsharansatsangi You're welcome. I hope that's what you meant by the override default sorting column.

sumitsharansatsangi:
Yeah @sinisaos , You have done exactly the same what I mean.

sinisaos:
@dantownsend No, that would be the default value where the first value in the select is `primary key`, but if you want the first value in the select to be `sort_column` if exists, we just need a small change by setting `sort_column` as a `propertyName` like this` propertyName: this.schema.sort_column_name`.
Both variants are fine with me, but if you want `sort_column` to be the default value, I can change that.

dantownsend:
@sinisaos I think that would be good.

dantownsend:
OK, just realised we also need to be able to specify ascending / descending.

dantownsend:
@sinisaos Thanks for fixing those tests.

sinisaos:
@dantownsend No problem :+1:

sinisaos (commented):
The solution without `time.sleep` is much better and the tests now pass every time locally for me, which was not the case before. Great

dantownsend:
@sinisaos It's a bit trickier writing tests for SPAs with Playwright, but those patterns seem to work OK.

dantownsend:
I just released a new version of Piccolo API which will let us sort by multiple columns: https://github.com/piccolo-orm/piccolo_api/releases/tag/0.52.0

I'll try and add support for it to this PR, so expose it in Piccolo Admin.

sinisaos:
I thought that one `sort_column` is enough, but fell free to do as you think is best. I tried the last Piccolo API release and sorry if I don't understand well (or doing something wrong) but order of result rows is the same with `GET /?__order=-rating,name` or `GET /?__order=-rating`? Is that the right behavior?

dantownsend:
> I thought that one sort_column is enough, but fell free to do as you think is best. I tried the last Piccolo API release and sorry if I don't understand well (or doing something wrong) but order of result rows is the same with GET /?__order=-rating,name or GET /?__order=-rating? Is that the right behavior?

It depends on the data. So if they all have different ratings, then the second column will make no difference. Set a bunch of them to have the same rating, and then try again - the ones with the same rating will sorted by name.

dantownsend:
> I thought that one sort_column is enough

Most of the time a single column is enough, but the project I'm using Piccolo on at the moment has some massive tables, where sorting by multiple columns is really useful. I've made some progress on it.

sinisaos:
Now I understand. If some rating of the movie is the same, if we give another param that rows with same rating will be sorted by name. Something like this. With `GET /?__order=-rating,name` results is
```json
{
    "rows":[
        {
            "id":1,
            "name":"A",
            "rating":10
        },
        {
            "id":2,
            "name":"B",
            "rating":5
        },
        {
            "id":3,
            "name":"C",
            "rating":5
        },
        {
            "id":4,
            "name":"D",
            "rating":5
        },
        {
            "id":5,
            "name":"E",
            "rating":1
        }
    ]
}
```
but with `GET /?__order=-rating,-name` result is 
```json
{
    "rows":[
        {
            "id":1,
            "name":"A",
            "rating":10
        },
        {
            "id":4,
            "name":"D",
            "rating":5
        },
        {
            "id":3,
            "name":"C",
            "rating":5
        },
        {
            "id":2,
            "name":"B",
            "rating":5
        },
        {
            "id":5,
            "name":"E",
            "rating":1
        }
    ]
}
```
Is that correct or am I missing the point?

dantownsend:
@sinisaos yes, exactly

sinisaos:
In that case, that's great. Thank you for the explanation.

dantownsend:
@sinisaos Have a look at this when you have a chance. I think it's finally done.

If the user specifies `order_by` in `TableConfig`, then the UI uses that. The user can override it in the modal like usual, but can sort by multiple columns:

Screenshot 2023-04-16 at 21 30 40

The sort is now persisted in the URL:

Screenshot 2023-04-16 at 21 31 27

sinisaos:
@dantownsend This is great. This has become a big PR from the relatively simple `sort_column` feature to multi-column sorting. I hope people will use this feature to reward your great effort on this.

sumitsharansatsangi:
@dantownsend , @sinisaos , Multi-column sorting is indeed a Great Feature. Thanks for implementing it , @dantownsend .

dantownsend:
@sinisaos @sumitsharansatsangi Thanks! I'll definitely find this useful in my own projects. Just a few tests to write and should be finished.

dantownsend:
I've added a little indicator to the table to show which columns are being ordered by. If you click the little triangle it will open the sort modal.

Screenshot 2023-04-17 at 11 34 52

Issue #271, opened 2022-12-28
Cannot edit non-default primary key in admin


PawelRoman:
Hi,

I have a model where a PK is defined on an explicit Varchar field, like this

```
class MyModel(Table):
    identifier = columns.Varchar(length=64, primary_key=True)
   .....  other fields .....
```

It's entirely on purpose. I don't want the PK to be an auto-incremented integer. I want a random string as PK. ORM allows for that, which is good.

The problem is trying to add those objects via admin. Admin UI does not show any textbox to enter the PK value and it is not possible to add those objects.  

Apparently admin assumes that primary key column is always an auto-incremented integer. But ORM allows to have PK on non-int non-incrementable column.

Any advice how to work around this problem?

sinisaos:
@PawelRoman Piccolo Admin does not provide a primary key column in forms. The primary key is automatically generated. To achieve what you want, you need to provide a random string as the default value like this
```python
import random
import string

def random_string():
    return "".join(random.choices(string.ascii_lowercase, k=64))

class MyModel(Table):
    identifier = Varchar(
        default=random_string,
        length=64,
        primary_key=True,
    )
```
Hope this helps.

PawelRoman:
Sorry, I kind of simplified it. I don't want any defaults. I want the field to be explicitly required. The very nature of this field is that it must be supplied by the caller on creating object (doesn't matter if creating it programmatically or in admin). It just cannot be happily generated as random string by the back-end. 

In django, it'd do that by having null=False and no default defined.

sinisaos:
@PawelRoman Ok, but as I wrote earlier Piccolo Admin does not provide a primary key column in forms. The primary key is automatically generated because forms are generated from Pydantic schema and primary key is [excluded
](https://github.com/piccolo-orm/piccolo_api/blob/0bd38e446bb5c5b95c6c18ca53d88572c98f080b/piccolo_api/crud/endpoints.py#L262-L272) By removing the `exclude_columns` line (you can try patching your local Piccolo API installation), your column will be displayed in the form (`Serial` primary key remain hidden which is good), but such changes can only be approved @dantownsend.

BezBartek:
@sinisaos @PawelRoman 
Two thoughts:
1. Excluding of the primary key should be applied only if the primary key is auto-increment type, otherwise is always a problem.  I think that we should propose that as a new feature/fix (What is your opinion @dantownsend? Thank you :))
[...]

Issue #138, opened 2021-12-15
Add `read_only_columns` to `TableConfig`


dantownsend:
We could add a `read_only_columns` option to `TableConfig`. When set, the UI would then render that column's input field as read only in the edit page.

```python
TableConfig(Post, read_only_columns=[Post.created_on])
```

It's useful because sometime there's stuff you don't want users to be able to modify.

PR #289, opened 2023-03-01
add read_only_columns to TableConfig

sinisaos:
Related to #138 and partially to #271


dantownsend:
@sinisaos Thanks for this.

It looks like we're hiding inputs if they're marked as readonly, but another option is to set the `readonly` attribute, so they are still visible, just not editable:

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

You should be able to do this using Vue:

```
:readonly="isReadOnlyString(columnName)"
```

sinisaos:
> You should be able to do this using Vue:
> 
> ```
> :readonly="isReadOnlyString(columnName)"
> ```

@dantownsend I already tried that but that doesn't work. I think hidding field (to disable the editing of certain form fields) is not a bad solution because we have, in addition to input fields, various selects, DurationWidget, ChoiceWidget, FlatPickr, etc. where the `readonly` attribute does not work and we have to use the `disabled` attribute and we would have to change a lot more code.

sinisaos:
@dantownsend With the latest commit the field should be visible but `read-only` or `disabled` in the edit form. I hope this is better and I hope I have included all the possible fields.

dantownsend:
@sinisaos It's OK - I can run the tests locally. 

You can download the artefacts from GitHub actions which might provide some clues:

https://github.com/piccolo-orm/piccolo_admin/actions/runs/4321016381

sinisaos:
@dantownsend Once again, the Playwright proved to be very useful. I changed a lot of code to pass the test with custom forms because there was a conflict in the schemas. I tried using your approach. Can you check it when you have time?

dantownsend:
@sinisaos Yeah, the Playwright tests are great - I'd like to write more. I will add some for nullable columns soon.

sinisaos:
@dantownsend You don't even have to write tests, you can generate tests with [playwright codegen](https://playwright.dev/python/docs/codegen) and then copy-paste the generated code.

dantownsend:
@sinisaos I don't mind the auto generated tests. When there are lots of tests I like to follow this approach:

https://playwright.dev/docs/pom

piccolo_admin/endpoints.py
```diff
@@ -774,7 +790,7 @@ async def store_file(
 
         try:
             file_key = await media_storage.store_file(
-                file_name=file.filename,
+                file_name=file.filename,  # type: ignore
```
dantownsend, line 793: What's the error with this?

dantownsend:
This looks great. I just need to test it locally.

piccolo_admin/example.py
```diff
@@ -322,6 +322,10 @@ async def booking_endpoint(request, data):
         Movie.release_date,
     ],
     rich_text_columns=[Movie.description],
+    read_only_columns=[
+        Movie.release_date,
+        Movie.won_oscar,
+    ],
```
dantownsend, line 328: One thing we might want to do is create a new table just for testing read only columns. I did something similar for nullable columns:

https://github.com/piccolo-orm/piccolo_admin/blob/2beacfe3d0285fcf0e09a1141ad26121906a91f5/piccolo_admin/example.py#L213-L222

So we have a table with every possible column type, so we can quickly see that they all work. It'll also make it easier when writing Playwright tests.

piccolo_admin/endpoints.py
sinisaos: There was a mypy error for this, but now I tried to clone the branch into a fresh virtualenv, ran linter and the error went away. Sorry, it's my fault. I will remove it.

dantownsend: No problem - just thought it might be something I needed to fix.

piccolo_admin/example.py
sinisaos: That's a great idea. I will try to make that table with all possible column types and test it.

dantownsend:
Ideally this would be verified on the API level too, but could maybe implement that later on in `PiccoloCRUD`. For now, we might need to add something to the docs warning people that it's the case.

piccolo_admin/example.py
dantownsend: Thanks

dantownsend:
I would put the `ReadOnlyColumns` table also under the `Testing` group.

Screenshot 2023-03-06 at 21 26 39

dantownsend:
I've tested it. I noticed a couple of things.

Array columns still let you add or remove items:

Screenshot 2023-03-06 at 21 30 15

Read only fields only work on the edit page. So this is a tricky one ... should they also be read only when adding a new row? This means the value will be the default. I think it's useful for columns like the primary key, created on / modified on etc.

dantownsend:
Another suggestion - we could put `read only` next to the label, so it's obvious that it's read only. Otherwise users might think it's just broken.

Screenshot 2023-03-06 at 21 40 16

You can add this CSS to show a special cursor too:

```css
input:read-only {
    cursor: not-allowed !important;
}
```

sinisaos:
@dantownsend I added a read-only to field label and disabled links for `ArrayWidget` in read-only mode.

> Read only fields only work on the edit page. So this is a tricky one ... should they also be read only when adding a new row? This means the value will be the default. I think it's useful for columns like the primary key, created on / modified on etc.

That was my intention because it was required in the #271. You can easily enable read-only mode in add form as well by removing `this.$route.params.rowID !== undefined` from the `isReadOnly` method in `RowForm.vue` but I don't think it's necessary because the default value is given by default and you can change it or not, and read-only columns are usually to prevent data editing. Sorry if I missed your point, but feel free to change it, if you think that's better.

Issue #298, opened 2023-04-26
Custom Forms List Display


Fidel-C:
If a custom form field is set to a list, it displays an add icon button which seems not to function as it should.

Here is my code for instance

```python
class EmailForm(BaseModel):
    subject: str = 'Booking'
    sender: str = 'Booking'
    receivers: list[EmailStr]
    body: str
    
async def emailform_endpoint(request, data: EmailForm):
    message=MessageSchema(
        subject=data.subject,
        subtype=MultipartSubtypeEnum.mixed,
        recipients=data.receivers,
        body=data.body
    )
```

The add icon button tied to the receivers field does not display an email input when clicked. It probably just appends an empty string to the list.

sinisaos:
@Fidel-C Yes, you're right. It's a small fix. I will make PR for that.

Fidel-C:
_Thanks. Looking forward to the fix._

dantownsend:
OK, I see the problem. I put this simple app together which replicates the problem:

```python
from pydantic import BaseModel
from piccolo_admin.endpoints import create_admin, FormConfig
from piccolo.apps.user.tables import BaseUser
from piccolo_api.session_auth.tables import SessionsBase
from piccolo.engine.sqlite import SQLiteEngine
from starlette.requests import Request


DB = SQLiteEngine()


class Model(BaseModel):
    items: list[str]


async def emailform_endpoint(request: Request, data: Model) -> str:
    return 'ok'


class BaseUserCustom(BaseUser, db=DB):
    pass


class SessionsBaseCustom(SessionsBase, db=DB):
    pass


app = create_admin(
    tables=[],
    forms=[
        FormConfig(
            name='Demo',
            pydantic_model=Model,
            endpoint=emailform_endpoint
        )
    ],
    auth_table=BaseUserCustom,
    session_table=SessionsBaseCustom
)

async def main():
    await BaseUserCustom.create_table(if_not_exists=True)

    if not await BaseUserCustom.exists().where(
        BaseUserCustom.username == 'piccolo'
    ):
        await BaseUserCustom.create_user(
            username='piccolo',
            password='piccolo123',
            admin=True,
            active=True
        )

    await SessionsBaseCustom.create_table(if_not_exists=True)


if __name__ == '__main__':
    import asyncio
    asyncio.run(main())

    import uvicorn
    uvicorn.run(app)
```

[...]

PR #299, opened 2023-04-26
enable list items in custom forms

sinisaos:
Small fix related to enable item list display in custom forms. Related to #298


dantownsend:
@sinisaos Thanks for this. I don't fully understand what's going on yet, so need to try replicating the issue locally.

dantownsend:
@sinisaos You're right - this is the fix. Thanks 👍

Issue #307, opened 2023-05-09
Allow custom links in the sidebar


dantownsend:
Discussed in https://github.com/piccolo-orm/piccolo_admin/discussions/306

We can add a new section in the sidebar containing user defined links:

![piccolo_admin_links_prototype](https://github.com/piccolo-orm/piccolo_admin/assets/350976/e89fb3a1-1e02-42fc-8e0f-1e1d6d1ee695)

The API would be pretty simple:

```python
app = create_admin(
    ...,
    sidebar_links=[
        ('Charts', '/custom-endpoint'),
        ('Top Movies', '/#/movie?__order=-box_office'),
        ...
    ]
)
```

It's useful for many reasons:

* In the `Top Movies` example, we're providing a quick way to get to certain pages with filters / sorting pre-applied.
* In the `Charts` example, this is if someone has added a custom page that they want to link to.

Considerations:

* Do we call it `links` or `shortcuts`?

Fidel-C:
I think 'shortcuts' is more descriptive.

sinisaos:
> I suppose that the sidebar links will be visible to only the admin users and not show up at all in the swagger docs?

@Fidel-C Piccolo Admin is protected by session auth and only the authenticated admin user can see the admin interface and admin Swagger docs. 
Want to do PR for this? 
If not, I can make a working prototype.

Fidel-C:
@sinisaos  alright thanks for the answer.

sinisaos:
@Fidel-C No problem. You didn't say anything about whether you will do PR or not? Do you want me to do initial PR?

Fidel-C:
> @Fidel-C No problem. You didn't say anything about whether you will do PR or not? Do you want me to do initial PR?

Sorry about that. I think you should.

sinisaos:
@Fidel-C PR #308, so you can try it if you want.

dantownsend:
@sinisaos Thanks

PR #308, opened 2023-05-11
add custom links option in sidebar nav

sinisaos:
Related to #307.


piccolo_admin/endpoints.py
```diff
@@ -401,6 +401,7 @@ def __init__(
         translations: t.List[Translation] = None,
         allowed_hosts: t.Sequence[str] = [],
         debug: bool = False,
+        sidebar_links: t.List[t.Tuple[str, str]] = [],
```
dantownsend, line 404: I think this was my fault, suggesting tuples. But if we just the user pass in a dict instead, we no longer need `get_sidebar_links`.

```
sidebar_links={
    "Top Movies": "/movies/?order=-popularity
```

dantownsend (commented):
This looks great - thanks!

Sorry it took me so long to review. Just one suggestion. And then we'll add docs and release it.

sinisaos:
@dantownsend I've updated the `sidebar_links` docstring and added an example. Is that enough because `create_admin` docs don't have a separate section for parameters, they are from the source code docstring?

piccolo_admin/endpoints.py
dantownsend: > Am I missing something?

No, you're right - I just meant we didn't need the bit that mapped tuples into dictionaries any more, as they're already dictionaries.

sinisaos:
> I see what you're saying. It might be nice to have a screen grab, so people get a better idea of what it does.
> 
> Based on the current docs, we would probably need a new page called `Sidebar Links` or something.

I agree. I can do it, but I think it would be better if you do it, because I see that almost all screenshots are from Safari web browser (and I don't have a Mac).

dantownsend:
> I agree. I can do it, but I think it would be better if you do it, because I see that almost all screenshots are from Safari and Mac (and I don't have a Mac).

OK, no problem.

Something like this might be OK, which doesn't have the border or address bar, so is browser / OS agnostic:

https://piccolo-admin.readthedocs.io/en/latest/_images/forms_sidebar.png

sinisaos:
If that's the case, I can try to make docs with screenshots and make a new commit.

dantownsend:
> If that's the case, I can try to make docs with screenshots and make a new commit.

Thanks!

dantownsend:
Made some minor changes:

* I rearranged the docs a bit - gave `Side Bar Links` its own page.
* Using normal `a` tags instead of `router-link`, so external links also work.

dantownsend:
@sinisaos If you're happy with the changes I made, will merge it in tomorrow.

sinisaos:
@dantownsend Your changes look great. Thanks. Feel free to merge it.

Issue #309, opened 2023-05-16
Small improvements to the login page


sinisaos:
Small changes to the login page by making the form stand out a bit and showing/hiding the password.

https://github.com/piccolo-orm/piccolo_admin/assets/30960668/64f4ac7e-dc7f-4efd-a53e-d3ce19ef1bed

Fidel-C:
Nice one.
I deployed a piccolo FastApi project to render.com, using ip 0.0.0.0 and port 10000. I created an admin user from terminal which is connected to a supabase db. But each time I try to login from the admin panel, I get "there's a problem loggin in". I also set `allowed_hosts[]` to my  render.com project url, still doesn't work. What could be wrong?

sinisaos:
> Nice one.

Thanks.

> I deployed a piccolo FastApi project to render.com, using ip 0.0.0.0 and port 10000. I created an admin user from terminal which is connected to a supabase db. But each time I try to login from the admin panel, I get "there's a problem logging in". I also set `allowed_hosts[]` to my render.com project url, still doesn't work. What could be wrong?

I've never used `render.com`, but I see it as an alternative to Heroku (especially after they discontinued the free plans). So you can connect Piccolo to the db, do migrations and create a user? You added `allowed_hosts` to` create_admin` without www or https like this `allowed_hosts=[your-app.render.com]`. What does the server log say about the error? Have you tried FastAPI in debug mode to see what happens? It's hard to say what the problem could be with too little information.
`

Fidel-C:
Thanks a lot for your help @sinisaos . I was able to resolve this issue by changing some things in my .env variables.


But the challenge now is attempting to login from the Vue front-end,  which also hosted separately. It continues to show `404 not found` in the render.com logs. I'll still figure that out hopefully.

sinisaos:
@Fidel-C Glad Piccolo Admin is working now. For a separate frontend Vue application, see the backend CORS settings to enable your backend API URL and axios setup (if you use axios) on frontend. When Heroku has a free plan, I remember serving the bundeled Vue app from the FastAPI static directory and everything was fine, something like this (just an example, I can't remember exactly)
```python
app = FastAPI()

app.mount('/', StaticFiles(directory='path-to-your-dist-folder/dist', html=True))
```
. This is an option if you cannot manage to serve the frontend app from a separate host. Good luck.

Fidel-C:
[...]

PR #310, opened 2023-05-23
Show/hide password on login page

sinisaos:
Related to #309


sinisaos:
> There's one thing which I'm a bit worried about though. When the password is saved by Chrome, someone can click the eye, and see the saved password. It seems like a security problem, because if someone gained quick access to a user's computer they could get their password.
> 
> But having said that, you can get it using JavaScript like this:
> 
> ```js
> document.querySelector('input[name=password]').value
> ```
> 
> So I'm not sure if we should have it on the login page - is it a security problem? I think it just highlights how problematic passwords are.

If someone got quick access to the user's computer, there's not much we can do (with or without show/hide password) because then it's easy to find the password with 
```js
document.querySelector('input[name=password]').value 
```
from the console, like you wrote or if the user has saved the password through the Chrome settings password manager.

If you're concerned about security risk, I'd close this PR as it's a minor (insignificant) change I've seen on many login forms and I don't think we should bother with it anymore. Do you agree with that?

dantownsend:
> If someone got quick access to the user's computer, there's not much we can do (with or without show/hide password) because then it's easy to find the password with

I agree that if someone got access to someone's computer they're basically screwed anyway.

> If you're concerned about security risk, I'd close this PR as it's a minor (insignificant) change I've seen on many login forms and I don't think we should bother with it anymore. Do you agree with that?

I wouldn't close this PR - I prefer the new header style on the login page. I'm probably being overly cautious with the view password thing, as even Google does this:

Screenshot 2023-05-24 at 10 20 32

I'd definitely like it on the change password page.

sinisaos:
> I wouldn't close this PR - I prefer the new header style on the login page. I'm probably being overly cautious with the view password thing, as even Google does this:

Yes exactly, even Google has it and I think they care about security. So we leave the current code? If yes, I will add new commit to show/hide password on the change password page.

sinisaos:
@dantownsend I made a new commit with the changes to the password change page. I only `show/hide` the new password confirmation. Is that enough or should we enable `show/hide` on all three form fields?

dantownsend:
I think we can make the password field as a separate component.

sinisaos:
@dantownsend That's a great idea. I will make another commit with those changes. I hope it will be good enough.

dantownsend:
@sinisaos Cool, thanks. I haven't used v-model on custom components much before, but seems like a clean solution.

https://vuejs.org/guide/components/v-model.html

sinisaos:
@dantownsend Great. This can be merged because everything works and looks good.

Issue #146, opened 2022-01-27
Chart support


dantownsend:
I'm putting this here as a placeholder to collect thoughts. I don't think it's an immediate priority, but would be cool.

We support form creation using `FormConfig`. We might be able to add something similar called `ChartConfig`. It would initially just be for very simple chart types, like bar, line, and pie.

Here's an example:

```python
async def get_data():
    movies = await Movie.select(
        Movie.director.name.as_alias("director"),
        Count(Movie.id)
    ).group_by(
        Movie.director
    )
    # Flatten the response so it's a list of lists like [['George Lucas', 3], ...]
    return [[i['director'], i['count']] for i in movies]

director_chart = ChartConfig(title='Movie count', type=Bar(source=get_data))

create_admin(charts=[director_chart])
```

There would be a new section in the sidebar:

<img width="240" alt="Screenshot 2022-01-27 at 14 24 09" src="https://user-images.githubusercontent.com/350976/151379999-d306b922-d2ec-44e4-b112-6954eb7e7bd2.png">

sinisaos:
@dantownsend @Fidel-C I've been playing around a bit with adding charts to Picolo Admin. In addition to `custom links`(#307), we can also have a separate Charts section. The result is this.

https://github.com/piccolo-orm/piccolo_admin/assets/30960668/1065f78d-8b7c-4139-a111-43317e862cac


The data format must be a `list of lists` (eg. `data=[["Male", 7], ["Female", 3]]`) to satisfy the frontend library (I used the [Chartkick](https://chartkick.com/vue) library). I used hardcoded data, but any function or coroutine which return list of lists, will satisfy the data format.

Fidel-C:
@sinisaos wow this is really nice

dantownsend:
@sinisaos Yeah, that's really cool 👍 

How large is the chartkick library? We just need to make sure we pick something which doesn't make the bundle size too large.

dantownsend:
I'm already thinking how I could use this in my own projects ... I think it'll be very useful.

Fidel-C:
@sinisaos is the chart feature now in the official version?

sinisaos:
[...]

PR #311, opened 2023-05-26
Charts support

sinisaos:
Related to #146. @dantownsend I created this PR so you can try it, change it and play with this interesting new feature.


admin_ui/package.json
```diff
@@ -18,13 +18,16 @@
...
+        "chart.js": "^4.3.0",
         "core-js": "^3.6.5",
         "flatpickr": "^4.6.9",
+        "highcharts": "^11.0.1",
         "is-svg": "^4.3.1",
         "js-cookie": "^2.2.1",
         "json-bigint": "^1.0.0",
         "ssri": "^8.0.1",
         "vue": "^2.7.14",
+        "vue-chartkick": "^0.6.1",
```
dantownsend, line 30: What's the interaction between `chart.js`, `highcharts` and `vue-chartkick`?

admin_ui/src/components/TableNav.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div>
+    <div class="sidebar_wrapper">
```
dantownsend, line 2: What's this for?

admin_ui/src/router.ts
```diff
@@ -8,6 +8,7 @@ import Login from './views/Login.vue'
 import ChangePassword from './views/ChangePassword.vue'
 import RowListing from './views/RowListing.vue'
 import AddForm from './views/AddForm.vue'
+import GetCharts from './views/GetCharts.vue'
```
dantownsend, line 11: Maybe we just call this view `Chart.vue`?

docs/source/charts/index.rst
```diff
@@ -0,0 +1,32 @@
+.. _Charts:
+
+Charts
+======
+
+Piccolo Admin can display different types of charts based on yours
```
dantownsend, line 6: Typo here: `yours` -> `your`.

docs/source/charts/index.rst
```diff
@@ -0,0 +1,32 @@
...
+data. Five chart types are supported: ``Pie``, ``Line``, ``Column``, 
+``Bar`` and ``Area``.
+
+Here's an example of charts usage, using FastAPI:
+
+.. literalinclude:: ./examples/app.py
+
+Piccolo Admin will then show a charts in the UI.
```
dantownsend, line 14: `charts` -> `chart`

docs/source/charts/index.rst
```diff
@@ -0,0 +1,32 @@
...
+
+.. image:: ./images/charts_sidebar.png
+
+.. image:: ./images/chart.png
+
+.. warning::
```
dantownsend, line 20: Maybe we make this a hint instead? so it's `.. hint::`?

docs/source/charts/index.rst
```diff
@@ -0,0 +1,32 @@
...
+
+  The data format must be a ``list of lists``
```
dantownsend, line 22: I'd remove the backticks here - it's usually just for quoting code.

docs/source/charts/index.rst
```diff
@@ -0,0 +1,32 @@
...
+Piccolo Admin will then show a charts in the UI.
+
+.. image:: ./images/charts_sidebar.png
+
+.. image:: ./images/chart.png
+
+.. warning::
+
+  The data format must be a ``list of lists`` 
+  (eg. ``[["Male", 7], ["Female", 3]]``).
```
dantownsend, line 23: `eg.` -> `e.g.`

piccolo_admin/endpoints.py
```diff
@@ -349,6 +349,60 @@ class FormConfigResponseModel(BaseModel):
...
+    :param title:
+        This will be displayed in the UI in the sidebar.
+    :param chart_slug:
+        This determines which chart will be displayed.
+    :param chart_type:
+        Available chart types. There are five types: ``Pie``, ``Line``,
+        ``Column``, ``Bar`` and ``Area``.
+    :param data:
+        The data to be passed to the admin ui. The data format must be
+        a ``list of lists`` (eg. ``[["Male", 7], ["Female", 3]]``).
```
dantownsend, line 366: ``list of lists`` -> list of lists

piccolo_admin/endpoints.py
```diff
@@ -349,6 +349,60 @@ class FormConfigResponseModel(BaseModel):
...
+        director_chart = ChartConfig(
+            title='Movie count',
+            chart_type="Pie", # or Bar or Line etc.
+            data=director_movie_count,
+
+        create_admin(charts=[director_chart])
+
+    """
+
+    def __init__(self, title: str, chart_type: str, data: t.List[t.Any]):
```
dantownsend, line 392: This is defined as a dataclass - so we can drop the `@dataclass` decorator and just keep the `__init__` method. With our docs, Sphinx handles normal Python classes better than dataclasses.

piccolo_admin/endpoints.py
```diff
@@ -349,6 +349,60 @@ class FormConfigResponseModel(BaseModel):
...
+        self.chart_slug = self.title.replace(" ", "-").lower()
+        self.chart_type = chart_type
+        self.data = data
+
+
+class ChartConfigResponseModel(BaseModel):
+    title: str
+    chart_slug: str
+    chart_type: str
+    data: t.List[t.Any]
```
dantownsend, line 403: If the data returned a list of tuples ... we could define it more precisely like:

```
data: t.List[t.Tuple[str, t.Union[int, float, decimal.Decimal]]]
```

Presumably each element is a string, followed by some kind of number.

piccolo_admin/endpoints.py
```diff
@@ -845,6 +920,38 @@ def get_user(self, request: Request) -> UserResponseModel:
             user_id=str(request.user.user_id),
         )
 
+    ###########################################################################
+    # Custom charts
+
+    def get_charts(self) -> t.List[ChartConfigResponseModel]:
+        """
+        Returns  all charts registered with the admin.
```
dantownsend, line 928: I think we have an extra space here -> `Returns all`

piccolo_admin/endpoints.py
```diff
@@ -845,6 +920,38 @@ def get_user(self, request: Request) -> UserResponseModel:
...
+                chart_slug=chart.chart_slug,
+                chart_type=chart.chart_type,
+                data=chart.data,
+            )
+            for chart in self.charts
+        ]
+
+    def get_single_chart(self, chart_slug: str) -> ChartConfigResponseModel:
+        """
+        Returns single chart.
```
dantownsend, line 942: -> `Returns a single chart`

piccolo_admin/endpoints.py
dantownsend: I think the type annotation for data is wrong. In the docstring it shows that we can pass in an async function.

piccolo_admin/example.py
```diff
@@ -442,6 +443,22 @@ def booking_endpoint(request: Request, data: BookingModel) -> str:
...
+                ["Peter Jackson", 6],
+                ["Ron Howard", 1],
+            ],
+        ),
+        ChartConfig(
+            title="Director gender",
+            chart_type="Column",
+            data=[["Male", 7], ["Female", 3]],
+        ),
+    ],
```
dantownsend, line 461: I'd add examples for each chart we support - I think we support Pie, Line, Column, Bar and Area.

Also, I think the main missing piece if the data is hardcoded at the moment, but in the docs we show how the user can pass in an async function, so the data can be fetched dynamically.

dantownsend (commented):
Very interesting! I left some comments. I think it's mostly done - the main thing at the moment is getting the chart data to be returned dynamically.

dantownsend:
A couple of thoughts on the this screen:

Screenshot 2023-06-28 at 00 53 38

Ideally the chart would take up more space if available.

Also, one some pages in Piccolo Admin we don't show the sidebar. I think for the charts it would be better to see the sidebar, as it makes it easier to jump between charts.

admin_ui/package.json
dantownsend: Is `chart.js` redundant? It looks like we're just using `highcharts`?

dantownsend: Have you tried `chart.js`? It's slightly smaller then highcharts, and it supports tree shaking, so we might get a smaller bundle size:

https://bundlephobia.com/package/chart.js@4.3.0
https://bundlephobia.com/package/highcharts@11.1.0

Also, the licensing on highcharts looks complicated, so I'd rather stay away from it. chart.js is MIT like Piccolo Admin.

https://github.com/chartjs/Chart.js/blob/master/LICENSE.md

sinisaos: My mistake. I tried both libraries, so it accidentally stayed in `package.json`. This PR uses `highcharts`, but I will switch to `Chart.js` in the next commit.

admin_ui/src/components/TableNav.vue
sinisaos: This is for a menu sidebar of equal width between the `Home` page and other pages. Now this is not the case because the sidebar menu on the `Home` page is wider than on other pages. I can remove it if you want?

admin_ui/src/router.ts
sinisaos: I will change that.

piccolo_admin/endpoints.py
sinisaos: No, we can pass **the result** of async or sync function which is a `list of lists`. I will change the annotation to `t.List[t.List[t.Any]]`. I will also update the docstring.

piccolo_admin/endpoints.py
sinisaos: Returned data must be `list of lists`. I will change the annotation to `t.List[t.List[t.Any]]`.

piccolo_admin/example.py
sinisaos: Yes, you're right, the data is hardcoded because we can't run database queries in the piccolo admin example (`piccolo_conf` error), but we can use a coroutine or sync function in a normal application. I will change the test to run the hardcoded data from the coroutine in `example.py` so that the user can see an example of how to use it. Something like this
```python
async def fetch_chart_data() -> t.List[t.List[t.Any]]:
    return [
        ["George Lucas", 4],
        ["Peter Jackson", 6],
        ["Ron Howard", 1],
    ]

chart_data = asyncio.run(fetch_chart_data())

APP = create_admin(charts=[
        ChartConfig(
            title="Movie count Pie",
            chart_type="Pie",
            data=chart_data,
        ),
        ChartConfig(
            title="Movie count Line",
            chart_type="Line",
            data=chart_data,
        ),
        ChartConfig(
            title="Movie count Column",
            chart_type="Column",
            data=chart_data,
        ),
        ChartConfig(
            title="Movie count Bar",
            chart_type="Bar",
            data=chart_data,
        ),
        ChartConfig(
            title="Movie count Area",
            chart_type="Area",
            data=chart_data,
        ),
  ],
)
```

piccolo_admin/endpoints.py
dantownsend: But if you just pass the result, the chart is no longer dynamic. I think each time the chart is viewed the data should be calculated from the function.

piccolo_admin/example.py
dantownsend: I think the problem is the query is being run before the engine has been assigned to the tables.

If we make it dynamic then we won't have this problem.

```python
async def fetch_chart_data() -> t.List[t.List[t.Any]]:
    return await Movie.select() ...

APP = create_admin(charts=[
        ChartConfig(
            title="Movie count Pie",
            chart_type="Pie",
            source=fetch_chart_data,
        ),
]
```

Then every time the chart is fetched from the API, we run the function to get the latest data.

dantownsend: I can make the change if you like - it's up to you.

sinisaos: Of course. That would be great. In the last commit I made your suggested changes. I think I covered everything except one thing I had trouble wrapping the chart pages in the `BaseView` (to have a sidebar) due to a weird Vue router warning 
`vue-router.esm.js:16 [vue-router] missing param for named route "chart": Expected "chartSlug" to be defined` even though the `chartSlug` is defined in the router like this
```javasript
{
    path: '/charts/:chartSlug',
    name: 'chart',
    component: Chart,
    props: true
},
```
so I left the `DetailBaseView`. It would be nice if you could look at that too. Thanks

dantownsend: OK, I'll create a branch, and will play around.

sinisaos:
Used in #316

PR #316, opened 2023-06-30
Chart experiments

dantownsend:
Based on https://github.com/piccolo-orm/piccolo_admin/pull/311

The only real change is the data is calculated dynamically each time the endpoint is called.


sinisaos:
@dantownsend This is great. You can use the [changes from the last commit](https://github.com/piccolo-orm/piccolo_admin/pull/311/commits/da9fe50cbd6450d7cc710fedc11617abd874b5b2) from the `Charts support` branch because only `Chart.js` is used there and there are changes in chart size, typo fixes, etc.

dantownsend:
@sinisaos Good catch - I've updated it with your latest changes.

dantownsend:
@sinisaos I like the way chart.js looks.

Screenshot 2023-07-01 at 08 27 50

The only problem is dark mode where the labels are harder to see.

Screenshot 2023-07-01 at 08 27 56

Do you know if Chart.js supports dark mode, or we can configure the label colours somehow?

sinisaos:
@dantownsend I don't know, but there's probably some way to change it through css or something.

docs/source/charts/images/chart.png
sinisaos, line 1: Now I noticed that the image in the docs is wrong  (use `Highcharts` and not `Chart.js`), so if you can change that too.

dantownsend: @sinisaos OK will change that

dantownsend:
@sinisaos I had an idea ... will play around with it to see how easy it is. We can let the user specify a Pydantic model which is used to configure the chart.

For example:

```python
class ChartModel(BaseModel):
    genre: str | None = None

async def get_director_movie_count(model: ChartModel):
    query = (
        await Movie.select(
            Movie.director.name.as_alias("director"), Count(Movie.id)
        )
        .group_by(Movie.director)
        .order_by(Movie.director.name)
    )
    if model.genre:
        query = query.where(Movie.genre == model.genre)
    movies = await query()
    return [(i["director"], i["count"]) for i in movies]

APP = create_admin(
    charts=[
        ChartConfig(
            title="Movies per director",
            chart_type="Pie",
            data_source=get_director_movie_count,
        ),
    ]
)
```

On the chart page we then have a button which opens a modal containig a form based on the Pydantic model. In this case there's a text input where the user can specify a genre, and then the chart just shows the number of movies per director in that genre.

sinisaos:
@dantownsend That would be interesting. I don't know how complicated it is to implement because you also need to check if the user types a good term and maybe you should use a list of predefined terms (and then use the select dropdown) to prevent bad terms.

admin_ui/src/main.ts
```diff
@@ -45,6 +47,7 @@ Vue.filter("readable", function (value) {
 /*****************************************************************************/
 
 Vue.config.productionTip = false
+Vue.use(Chartkick.use(Chart))
```
sinisaos, line 51:
```suggestion
// charts font color and gridlines in dark mode
if (localStorage.getItem('darkMode')) {
    Chart.defaults.global.defaultFontColor = "#ffffff"
    Chart.defaults.scale.gridLines.color = "#ffffff"
}
```

sinisaos: This is a way to change Chart.js colors in dark mode, but unfortunately it doesn't work without refreshing the page. I'm sorry I didn't check before posting the suggested change.

dantownsend: No worries, it's still useful - there might be a way to reload the chart without reloading the entire page.

dantownsend:
@sinisaos I got the chart filtering working:

https://github.com/piccolo-orm/piccolo_admin/assets/350976/2837451b-76af-4064-8c1c-ceb00b18dd3d

sinisaos:
@dantownsend Wow, that looks and works great and I think it's really useful.

dantownsend:
@sinisaos Thanks, I think it's pretty cool.

dantownsend:
I have a bunch of changes for this - just need to push them up.

Issue #313, opened 2023-06-22
Pydantic V2


sinisaos:
Since the Piccolo ecosystem depends on Pydantic, the new `Pydantic V2` beta is out and libraries like FastAPI are [available for testing](https://github.com/tiangolo/fastapi/releases/tag/0.100.0-beta1). `V2` brings a lot of braking backwards compatibility changes and I wanted to see how much work we'll have once we switch to `V2`. I managed to change the code and tests in Piccolo ORM and Piccolo API and everything works (some `DeprecationWarning` are left). The most changes are in Piccolo Admin, as expected, because the schema has changed (`anyOf` keyword which we don't use etc.). Now the properties fields are arranged alphabetically (as well as Piccolo Admin form fields which is a bit strange), but I see that this has changed and that everything will be like in `V1` (this [merged PR](https://github.com/pydantic/pydantic/pull/6212)). Mostly everything works except:

- <s>`DurationWidget` (Interval)</s> - **fixed**
- <s>`FlatPickr` (datetime widget)</s> - **fixed**
- <s>`ArrayWidget` (arrays and file-multifile upload)</s> - **arrays works** :smile: 
- <s>`nullable columns`</s> - **fixed**

mostly due to changes in schema. I'm sorry for the long comment, but I hope this will be useful and a placeholder for later changes we will need to make when we need to move to `V2`.

dantownsend:
@sinisaos Thanks for looking into it. It's a shame about the breaking changes to the schema - but we should be able to work around it. I wonder if the schema has some kind of version number in it.

sinisaos:
@dantownsend When the time comes, we'll probably be able to fix everything. There is now an `anyOf` keyword in the schema and it contains the formats if they are not defined in Piccolo [pydantic.py](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/utils/pydantic.py). The `textarea` is fine because 
```python
elif isinstance(column, Text):
    field = pydantic.Field(format="text-area", extra=extra, **params)
```
![textarea](https://github.com/piccolo-orm/piccolo_admin/assets/30960668/833b8fb2-30da-41ea-bc93-05c2b0029b94)

but for example `date` format is in the `anyOf` keyword and the Piccolo admin cannot currently read it from the schema.

![datetime](https://github.com/piccolo-orm/piccolo_admin/assets/30960668/c3353149-2ba1-4b34-b4b4-e67637bcd7ef)

[...]

PR #317, opened 2023-07-07
pin FastAPI and Pydantic 1 version

sinisaos:
Related to [this issue](https://github.com/piccolo-orm/piccolo_admin/issues/313)


dantownsend:
Going to pin FastAPI a bit looser, just in case.

PR #327, opened 2023-07-27
pydantic v2 support

sinisaos:
Related to [https://github.com/piccolo-orm/piccolo_admin/issues/313](https://github.com/piccolo-orm/piccolo_admin/issues/313).


sinisaos:
@dantownsend Yesterday all tests and linters were fine, but today nothing passed because `503 (Service Unavailable)` from Github. When you push another commit everything should be fine.

admin_ui/src/utils.ts
```diff
@@ -104,34 +125,44 @@ export function convertFormValue(params: {
 
     if (value == "null") {
         value = null
-    } else if (schema?.properties[key].type == "array") {
+    } else if (schema.properties[key]["anyOf"][0].type == "array") {
```
dantownsend, line 128: Ideally we could do with some kind of function for doing this match. It's unlikely to happen but if the element we need isn't at the first index it would fail.

Something like:

```javascript
const hasType(property, type: string): boolean => {
    const types = property.anyOf.map((i) => i.type)
    return types.indexOf(type) != -1
}
```

dantownsend: Hmm, some of them have three potential types:

Screenshot 2023-09-14 at 22 40 15

dantownsend: If they're in order, then maybe getting the first element is the way to go.

I wonder what happens if we have a Piccolo column with `required=True` - does it still return `anyof`?

sinisaos: No. If `required=true`, not `anyOf` is returned.

sinisaos: > Ideally we could do with some kind of function for doing this match. It's unlikely to happen but if the element we need isn't at the first index it would fail.
> 
> Something like:
> 
> ```js
> const hasType = (property, type: string): boolean => {
>     const types = property.anyOf.map((i) => i.type)
>     return types.indexOf(type) != -1
> }
> ```

In fact, for all columns whose `type` or `format` defined in Piccolo ORM [pydantic.py](https://github.com/piccolo-orm/piccolo/blob/4f898c1a2a971feb7bd77e36bfe7fc14dd7a5108/piccolo/utils/pydantic.py#L300-L353), we don't need `anyOf`. I will make a new commit with those changes. If it's not good, feel free to change it as you think is best.

admin_ui/src/utils.ts
```diff
@@ -106,32 +127,43 @@ export function convertFormValue(params: {
...
+        && value == "") {
         value = null
-    } else if (schema?.properties[key].type == "integer" && value == "") {
+    } else if (schema?.properties[key].format == "json" &&
+        schema?.properties[key].extra["nullable"] == true
+        && value == "") {
         value = null
-    } else if (schema?.properties[key].type == "number" && value == "") {
+    }
+    else if (schema?.properties[key]["anyOf"][0].type == "integer" &&
```
dantownsend, line 151: I wonder if it's possible to dramatically simplify this logic. I wonder if there are any column types left now, which when nullable and have an empty string as a value aren't converted to `null`?

sinisaos: Maybe we can use a few variables to reduce code repetition. Something like this
```javascript
export function convertFormValue(params: {
    key: string
    value: any
    schema: Schema
}): any {
    let { key, value, schema } = params

    const columnTypeAnyOf = schema?.properties[key]["anyOf"][0].type
    const columnFormat = schema?.properties[key].format
    const columnNullable = schema?.properties[key].extra["nullable"] == true
    const possibleFormats = [
        "uuid",
        "email",
        "date-time",
        "date",
        "json",
        "integer",
        "number",
        "string"
    ]

    if (value == "null") {
        value = null
    } else if (schema?.properties[key].type == "array") {
        value = JSON.parse(String(value))
    } else if (possibleFormats.includes(columnFormat) &&
        columnNullable
        && value == "") {
        value = null
    } else if (possibleFormats.includes(columnTypeAnyOf) &&
        columnNullable
        && value == "") {
        value = null
    } else if (schema?.properties[key].extra.foreign_key == true &&
        columnNullable &&
        value == ""
    ) {
        value = null
    }
    return value
}
```

Issue #328, opened 2023-07-28
CSV export without dependencies


sinisaos:
Implement `CSV` export without external dependencies.

PR #329, opened 2023-07-28
Implement CSV export without external dependencies

sinisaos:
Related to #328.


admin_ui/src/components/CSVButton.vue
```diff
@@ -59,21 +55,25 @@ export default {
...
                     exportedRows.push(...response.data.rows)
                 }
-                return exportedRows
+                let csv: string = "data:text/csv;charset=utf-8,"
+                csv += [
+                    Object.keys(exportedRows[0]).join(";"),
+                    ...exportedRows.map((item) => Object.values(item).join(";"))
+                ]
+                    .join("\n")
+                    .replace(/(^\[)|(\]$)/gm, "")
```
dantownsend, line 64: What does this regex do?

admin_ui/src/components/CSVButton.vue
```diff
@@ -59,21 +55,25 @@ export default {
...
+                const data: string = encodeURI(csv)
+                const link: HTMLAnchorElement = document.createElement("a")
+                link.setAttribute("href", data)
+                link.setAttribute("download", `${this.tableName}.csv`)
+                link.click()
```
dantownsend, line 69: OK, I think I understand what's happening here.

Does this created link appear anywhere in the document? Should it be deleted afterwards. Or just have a permanent hidden link, which we set the download and href attributes on using Vue.

admin_ui/src/components/CSVButton.vue
sinisaos: I found it in one of my old PRs (copy from internet) and it's a good question :). Square brackets should be escaped with that regex, but we don't need that at all. I just tried it now and the exported file looks the same without that line. Sorry I didn't think of that before. I will remove that line. Sorry.

admin_ui/src/components/CSVButton.vue
sinisaos: Yes, this creates a link with data and download attributes and the behavior is the same as in `vue-json-excel`. Feel free to change this as you think is best or we can always go back to using an external library.

dantownsend:
Merged in via https://github.com/piccolo-orm/piccolo_admin/pull/334

PR #376, opened 2024-03-12
Enhance ASGI Example in README.md

jrycw:
The original example is missing the imports for `Mount` and `Hello`. I propose using the [provided example for the documentation](https://github.com/piccolo-orm/piccolo_admin/blob/master/docs/source/asgi/examples/starlette/app.py) to replace it.


dantownsend:
@jrycw I agree, this is much better - thanks!

PR #205, opened 2022-08-28
Add experimental `zh-CN` translation to interface

mnixry:
Some words look hard to localization. (I don't have much experience in Chinese localization)

Any comments or change request are welcomed.

PR #378, opened 2024-03-16
Improve `zh-TW` translation

jrycw:
As mentioned in this [PR](https://github.com/piccolo-orm/piccolo_admin/pull/205), some words may not directly translate from English to either `zh-CN` or `zh-TW`, but we've made our best effort to provide accurate translations. :smile:


dantownsend:
@jrycw Great - thanks! I can imagine that translating it is very challenging.

PR #379, opened 2024-03-21
disable unique column in BulkUpdate dropdown

sinisaos:
Related to this [issue](https://github.com/piccolo-orm/piccolo_api/issues/273). This PR will disable unique columns in bulk update. This change will work when a new version of Piccolo ORM is released. Playwright tests will also pass after new Piccolo ORM release, because for now `BulkUpdate` dropdown is empty

### DIFF
--- a/admin_ui/src/components/BulkUpdateModal.vue
+++ b/admin_ui/src/components/BulkUpdateModal.vue
@@ -11,7 +11,10 @@
                     <option
                         :key="columnName"
                         :value="columnName"
-                        v-if="columnName != schema.extra.primary_key_name"
+                        v-if="
+                            columnName != schema.extra.primary_key_name &&
+                            property.extra.unique == false
+                        "
                     >
                         {{ property.title }}
                     </option>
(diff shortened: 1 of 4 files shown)


dantownsend:
Thanks for this - looks great.

When I've released a new version of Piccolo, we should update the minimum version of Piccolo in requirements.txt, and this should be good to go.

sinisaos:
@dantownsend It's weird, the Playwright tests should pass even without the latest Piccolo ORM version because the tested `gender column` in the `Director` table **is not a unique column**. Locally, all tests pass. I hope that after the new version of Piccolo everything will be fine.

dantownsend:
@sinisaos With the tests Playwright failing, it could either be that in the CI it's slower, and more error prone. Also, make sure you run `npm run build` before running the Playwright tests locally, just in case that's the problem.

sinisaos:
@dantownsend Ah, you're right. I forgot to run `npm run build` and the tests were passing locally because they were working with the old build. Sorry about that, but I haven't done PR for Piccolo Admin in a while. With the new build the `bulk_update` test failed just like in CI. If I patch the local Piccolo (add a `unique` column to the column extra) the test passes locally, which means the test will pass when we update the minimum version of Piccolo, as you said earlier. Thanks.

dantownsend:
@sinisaos Cool - seems to work now! Thanks

PR #389, opened 2024-05-02
use `convertFormValue` in custom forms

dantownsend:
Just a small change to reduce code repetition.


dantownsend:
Blocked by being unable to upload to codecov.io - trying a fix here: https://github.com/piccolo-orm/piccolo_admin/pull/390

Issue #386, opened 2024-04-25
Input field disappearing when clicking 'Add'.


AmazingAkai:
```py
class NewStaff(BaseModel):
    name: Annotated[str, Field(max_length=255, pattern=r"^[a-zA-Z\s]+$")]
    email: EmailStr
    superuser: bool
    permissions: list[Permission]
    
    
async def new_staff_endpoint(request: Request, data: NewStaff) -> str:
    ... # user creation code...


NEW_STAFF_FORM = FormConfig(
    name="New Staff",
    description="Create a new staff member.",
    pydantic_model=NewStaff,
    endpoint=new_staff_endpoint,
)
```
Permission is just an enum
```py
class Permission(Enum):
    UPLOADS = "uploads"
    NOTICES = "notices"
    GALLERY = "gallery"
    ADMISSIONS = "admissions"
```

[Screencast_20240425_213515.webm](https://github.com/piccolo-orm/piccolo_admin/assets/93782497/d98ad564-ef59-4d21-99cb-e7ab6c3eb837)

sinisaos:
@AmazingAkai This can be fixed by including optional parameter on this line 
https://github.com/piccolo-orm/piccolo_admin/blob/1157c120415b54e509f7720daaa587cd2cc9d450/admin_ui/src/components/ArrayWidget.vue#L118-L120

like this
```javascript
isMediaColumn() {
    return this.schema?.extra.media_columns.includes(this.fieldName)
}
```
Result is:

[add_button.webm](https://github.com/piccolo-orm/piccolo_admin/assets/30960668/4650cd76-07ac-47d0-b2ca-edad5b23ab2b)

But then we have another problem because Piccolo Admin doesn't understand `"$ref": "#/$defs/Permission" `(json reference) to `Permission` enum and enum attributes are not displayed as `HTML select`. It seems that custom forms can currently only work with `str`, `int`, `bool` etc. but not with json reference to enums or another Pydantic models.

dantownsend:
@sinisaos I think adding this for now makes sense:

```js
isMediaColumn() {
    return this.schema?.extra.media_columns.includes(this.fieldName)
}
```

At some point we need to refactor things a bit so `InputField` doesn't try getting the schema from the store, but I think this solution is good for now.

sinisaos:
[...]

PR #391, opened 2024-05-03
enable the use of Python enums in custom forms

sinisaos:
Related to #386

### DIFF
--- a/piccolo_admin/example.py
+++ b/piccolo_admin/example.py
@@ -449,7 +449,7 @@ def custom_form_choices() -> t.Dict[str, t.Any]:
 
 
 class NewStaff(BaseModel):
-    name: t.Annotated[str, Field(max_length=255, pattern=r"^[a-zA-Z\s]+$")]
+    name: str
     email: EmailStr
     superuser: bool
     permissions: t.Any = Field(
--- a/piccolo_admin/example.py
+++ b/piccolo_admin/example.py
@@ -48,7 +48,7 @@ from piccolo.table import Table, create_db_tables_sync, drop_db_tables_sync
 from piccolo_api.media.local import LocalMediaStorage
 from piccolo_api.media.s3 import S3MediaStorage
 from piccolo_api.session_auth.tables import SessionsBase
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, EmailStr, Field, field_validator
 from starlette.requests import Request
 
 from piccolo_admin.endpoints import (
@@ -430,6 +430,38 @@ def booking_endpoint(request: Request, data: BookingModel) -> str:
     return "Booking complete"
 
 
+# An example of using Python enum in custom forms
+class Permission(str, enum.Enum):
+    UPLOADS = "uploads"
+    NOTICES = "notices"
+    GALLERY = "gallery"
+    ADMISSIONS = "admissions"
+
+
+def custom_form_choices() -> t.Dict[str, t.Any]:
+    choices = {}
+    for item in Permission:
+        choices[item.name] = {
+            "display_name": item.name,
+            "value": item.value,
+        }
+    return choices
+
+
+class NewStaff(BaseModel):
+    name: t.Annotated[str, Field(max_length=255, pattern=r"^[a-zA-Z\s]+$")]
+    email: EmailStr
+    superuser: bool
+    permissions: t.Any = Field(
+        json_schema_extra={"extra": {"choices": custom_form_choices()}}
+    )
+
+
+def new_staff_endpoint(request: Request, data: NewStaff) -> str:
+    print(data)
+    return "A new staff member has been successfully created."
+
+
 TABLE_CLASSES: t.Tuple[t.Type[Table], ...] = (
     Director,
     Movie,
@@ -596,6 +628,12 @@ APP = create_admin(
             pydantic_model=BookingModel,
             endpoint=booking_endpoint,
             description="Make a booking for a customer.",
+        ),
+        FormConfig(
+            name="New Staff",
+            description="Create a new staff member.",
+            pydantic_model=NewStaff,
+            endpoint=new_staff_endpoint,
         ),
     ],
     auth_table=User,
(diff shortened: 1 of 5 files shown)


piccolo_admin/example.py
```diff
@@ -430,6 +430,38 @@ def booking_endpoint(request: Request, data: BookingModel) -> str:
...
+    return choices
+
+
+class NewStaff(BaseModel):
+    name: str
+    email: EmailStr
+    superuser: bool
+    permissions: t.Any = Field(
+        json_schema_extra={"extra": {"choices": custom_form_choices()}}
+    )
```
dantownsend, line 457: I think this is a clever workaround.

My concern though is it exposes a lot of internals.

How about modifying `FormConfig` instead? We could add an argument called `choices`.

```python
FormConfig("My Form", MyModel, choices={"some_field": SomeEnum})
```

We could do something similar with `help_text` etc.

Or we could give the user the option of passing in a Pydantic model, or a custom class like this:

```python
class MyModel(PiccoloAdminForm):
    name = Field(str, choices=SomeEnum, help_text="...")

FormConfig("My Form", MyModel, choices={"some_field": SomeEnum})
```

Which we then turn into a Pydantic model.

sinisaos: I was just trying to mimic the Piccolo column [choices](https://github.com/piccolo-orm/piccolo/blob/919983ac211ebf0d5953b126f6126b57deddb4cb/piccolo/columns/base.py#L236-L239) (just converting the enum to a dict with a `display_name` and a `value` that the admin understands how to render) because the Piccolo Admin doesn't understand references through `"$ref"` which is usually used to represent an enum or other complex type inside a Pydantic model (which was asked in the issue). I'm sorry, but I can't say that I fully understand and know how to implement your suggested changes. Feel free to implement the changes you think are best.

dantownsend: I was thinking we would use your logic, but rather than making the user do it, instead it would be done automatically by `FormConfig` in the `__init__`.

Something like:

```python
class FormConfig:
    def __init__(model, choices: dict, ...):
        for field_name, enum in choices.values():
            model.fields['field_name'].json_schema_extra["extra"]["choices"] = ...
```

The code above is pseudo code - I haven't tested it.

It would be used like this:

```python
FormConfig(MyModel, choices={"my_column": MyEnum"})
```

sinisaos: I tried your suggestion. I was able to dynamically add a `choices` field to a Pydantic model like this
```python
def convert_enum_to_choices(enum_data: t.Any) -> t.Dict[str, t.Any]:
    choices = {}
    for item in enum_data:
        choices[item.name] = {
            "display_name": item.name,
            "value": item.value,
        }
    return choices

class FormConfig:
    def __init__(pydantic_model, choices: t.Optional[t.Dict[str, t.Any]] = None, ...):
        self.pydantic_model = pydantic_model
        self.choices = choices
        if choices is not None:
            for field_name, field_value in choices.items():
                # update annotations
                pydantic_model.__annotations__.update({field_name: t.Any})
                # update model_fields
                pydantic_model.model_fields[field_name]: t.Any = Field(
                    json_schema_extra={
                        "extra": {
                            "choices": convert_enum_to_choices(field_value)
                        }
                    },
                )
                # update model_field annotation
                pydantic_model.model_fields[field_name].annotation = t.Any
                # if we print the fields, they are the same as when we define the permissions field in the NewStaff model as in example.py
                print(pydantic_model.fields) 
```
but the `choices` select is not displayed in the frontend form.

dantownsend: Does choices not appear in the `schema` in the endpoint?

sinisaos: No. The `choices` are not in the `schema` endpoint, so they cannot be displayed in the custom form interface either. Everything should work, but it doesn't.

sinisaos: I had to use [model_rebuild](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_rebuild) for the changes (dynamically adding `choices` in the Pydantic model) to take effect. Everything works now. The code is like this
```python
class FormConfig:
    def __init__(pydantic_model, choices: t.Optional[t.Dict[str, t.Any]] = None, ...):
        self.pydantic_model = pydantic_model
        self.choices = choices
        if choices is not None:
           for field_name, field_value in choices.items():
              # update model_fields
              pydantic_model.model_fields[field_name]: t.Any = Field(
                  json_schema_extra={
                      "extra": {
                          "choices": convert_enum_to_choices(field_value)
                      }
                  },
              )
              # update model_field annotation
              pydantic_model.model_fields[field_name].annotation = t.Any
              # rebuild the model for the changes to take effect
              pydantic_model.model_rebuild(force=True)
```

dantownsend: Good catch. Is it necessary to set the annotation? Or can we just keep what the user defined?

```python
pydantic_model.model_fields[field_name].annotation = str
```

sinisaos: We have to do this because otherwise the field annotation is `None` and validation fails when submitting the form. I made a new commit with the changes.

dantownsend: OK, I see. Can we use the annotation that the user originally set instead of `str`, because the `Enum` might be something else, like an `int`.

Do you think there's any change we'll be able to handle enums natively as they appear in OpenAPI? I haven't looked into it much, but this is the only caution I have around this now.

I definitely do want to use this approach for `help_text` though - I think that will be super useful.

sinisaos: > OK, I see. Can we use the annotation that the user originally set instead of `str`, because the `Enum` might be something else, like an `int`.
> 

I'm sorry, but I don't understand. The user does not set any annotation because the model field is dynamically added to the Pydantic model and we need to set the annotation otherwise the annotation is `NoneType`. If you mean the enum type we can use `any` type like this
```python
pydantic_model.model_fields[field_name].annotation = t.Any # type: ignore
```
but the result is the same as we use `str`.

> Do you think there's any change we'll be able to handle enums natively as they appear in OpenAPI? I haven't looked into it much, but this is the only caution I have around this now.

If you want to go with the native OpenAPI approach, that would be best, but Piccolo Admin or Piccolo API (`create_pydantic_model`) doesn't use complex references and for now uses choices which I try to mimic..

If you want to go with different approach I'll close this PR. I think this has become too complex. My first intention was to somehow allow users to use enums in custom forms (this is made possible by adding a `choices` to the `InputField`), but write custom code to use that enum. I gave an example of how to do it. After that, the `choices` argument is added to the `FormConfig`, with all the logic behind it, so the user doesn't have to do anything but just add the enum to the `choices` in the `FormConfig` constructor and it works. Sorry for the long comment.

Issue #396, opened 2024-06-11
Disabling the save button after click?


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo_admin/discussions/395

<div type='discussions-op-text'>

<sup>Originally posted by **mlauchli** June 11, 2024</sup>
I have a few hooks configured that fire when a record is updated. These submit an API request to a third party, and obviously the response back is not instantaneous. Is there a way to disable the save button upon click to ensure no repeat saves are made?</div>

PR #397, opened 2024-06-11
396 Disabling the save button after click

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_admin/issues/396


dantownsend:
I just thought, we should probably change the styles of the button when it's disabled, or show a spinner or something.

sinisaos (approved):
Sorry for the slow reply. I tried your changes on a slow network and everything works. The button is disabled until the API responds. The disabled button is a different color, but if you want you can increase the contrast even more with something like this
```css
button:disabled {
    background-color: #9fc8eb; /* or some other shade of color */
}
```

dantownsend:
@sinisaos Thanks. I did what you suggested - tried a few different colours, but in the end I reduced the opacity of the button a bit when disabled.

PR #405, opened 2024-08-16
MFA

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_api/pull/292

The workflow for MFA is pretty complex, so rather than reimplementing all of the UI in Vue, we'll just mount the MFA endpoint, and send the user there.

So the only bit of Vue code which needs changing is the login form, which needs to render an MFA code input.


piccolo_admin/endpoints.py
```diff
@@ -431,12 +433,17 @@ def __init__(
         allowed_hosts: t.Sequence[str] = [],
         debug: bool = False,
         sidebar_links: t.Dict[str, str] = {},
+        mfa_provider: t.Optional[MFAProvider] = None,
```
Skelmis, line 436: Bit of a cross reference without running it, is this intended to be typed as `t.Optional[t.Sequence[MFAProvider]]`? Or is the intention; for this release at least; to just provide support for setting up only a single provider?

I'm guessing based of the implementation at a glance that if someone were to create the relevant row entries in a table then they would still be able to authenticate it seems

dantownsend: I was going back and forth on it.

For an initial attempt, just supporting a single MFA provider is easier, but then it backs us into a corner if we want to later support multiple MFA providers.

I'll change it so it's a `t.Optional[t.Sequence[MFAProvider]]`. That means:

- Worst case scenario we'll just raise a `ValueError` for now if more than one is provider passed in, saying support for multiple MFA providers is coming soon.
- Or just support multiple MFA providers straight away - the amount of extra work is pretty minimal. We just need an extra UI view which lets the user select which MFA provider they're setting up.

Skelmis: Yea either choice is fine, just figured I'd call it out in the case it hadn't been considered. Possibly the following flow may help alleviate some of the needing to implement it pain while remaining future focused?
```python
class AdminRouter:
    def __init__(
        self,
        mfa_provider: t.Optional[t.Sequence[MFAProvider]] = None,
    ):
        self.mfa_providers: t.Optional[t.Sequence[MFAProvider]] = mfa_provider

        if self.mfa_providers:
            if len(self.mfa_providers) == 1:
                # Only one provider, skip the 'MFA selection' screen
                mfa_provider = self.mfa_providers[0]
                private_app.mount(
                    path="/mfa-setup/",
                    app=RateLimitingMiddleware(
                        app=mfa_setup(
                            provider=mfa_provider,
                            auth_table=self.auth_table,
                        ),
                        provider=rate_limit_provider,
                    ),
                )

            else:
                # More then one provider present, present
                # a selection screen for which one to set up
                raise ValueError(
                    "Piccolo does not currently support multiple MFA providers. This will be coming soon."
                )
```

dantownsend: Yes, exactly - that would work nicely.

piccolo_admin/endpoints.py
```diff
@@ -680,6 +687,21 @@ def __init__(
...
+            private_app.mount(
+                path="/mfa-setup/",
+                app=RateLimitingMiddleware(
+                    app=mfa_setup(
+                        provider=mfa_provider,
+                        auth_table=self.auth_table,
+                    ),
+                    provider=rate_limit_provider,
+                ),
+            )
```
Skelmis, line 703: Looks like the ratelimit provider is set to a more strict version if it's not provided in a few lines time. Can we also implement that here but with an even stricter ratelimit possible?

Some links from a reputable company adjacent to mine that popped up on my feed today as a throwback:
- Brute Forcing TOTP Multi-Factor Authentication is Surprisingly Realistic - [Link](https://pulsesecurity.co.nz/articles/totp-bruting)
- Supposing I want to brute force a numeric code? - [Link](https://totp.sova.st/)
```suggestion
        if mfa_provider:
               if not rate_limit_provider:
                   rate_limit_provider = InMemoryLimitProvider(
                       limit=5, timespan=60, block_duration=300
                   )

            private_app.mount(
                path="/mfa-setup/",
                app=RateLimitingMiddleware(
                    app=mfa_setup(
                        provider=mfa_provider,
                        auth_table=self.auth_table,
                    ),
                    provider=rate_limit_provider,
                ),
            )
```

dantownsend: Great catch!

Thanks for the link - it was an interesting read.

I've set the rate limiting to 5 requests in 5 minutes. But then I was thinking about it some more, and the `mfa-setup` endpoint is just for setting up MFA, and not for logging in. The `session_login` endpoint handles that, and uses the main `rate_limit_provider` - so might have to make the defaults on that stricter, or do some kind of account lock out.

Skelmis: Yea, something to consider anyway.

We should also likely change the session login so it's not a permanent rate limit by default

sinisaos:
@dantownsend I tried your changes and they are great, I really like them. Everything works and I really like your solution for the login form.

dantownsend:
@sinisaos Thanks! I think it's almost there now.

piccolo_admin/endpoints.py
dantownsend: For now, I'll raise a `ValueError` if more than one MFA provider is specified. I think we can support multiple providers in the future.

dantownsend:
I'm going to release this. There are some things which need improving - I added issues to cover these.

I'll try it out in some of my apps.

Thanks again both for all the input. 🙇‍♂️

sinisaos:
@dantownsend I don't know if that's a good idea, but can you add `MFA` related packages to requirements `extras`, just like in the Piccolo API. In the fresh env, an error is raised because packages related to MFA (`qrcode`, `pyotp`, `cryptography`, etc.) are missing. Thanks

dantownsend:
@sinisaos Yeah, I can see that being annoying.

The documentation needs to be quite a bit better too - will try and do that later today.

Issue #409, opened 2024-09-09
Make MFA Setup URL relative


dantownsend:
The URL for `MFA Setup` is problematic (`/api/mfa-setup/`):

https://github.com/piccolo-orm/piccolo_admin/blob/d787d7fe24828d54683dce0e3de6a75df6a7eb8b/admin_ui/src/components/NavDropDownMenu.vue#L17

If the user mounts Piccolo Admin at a sub URL, like `/admin`, the URL will return a 404. We need to make the URL relative (i.e. `./api/mfa-setup/`).

PR #410, opened 2024-09-09
409 Make `MFA Setup` URL relative

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_admin/issues/409


sinisaos:
@dantownsend Yes, that was a problem. I just wanted to write that it is necessary to make that changes because otherwise the link does not work properly, but you already make that changes. Great

sinisaos:
@dantownsend With that change everything works in my case.

Issue #414, opened 2024-09-29
Be able to download files via custom forms


dantownsend:
At the moment, custom forms just return a string, which is then displayed in Piccolo Admin.

It would be nice if custom forms could also return a `File` object which is then downloaded by the user.

This would be very useful if the custom form is being used to run a report, and the user wants to download a CSV file.

At the moment, the only choice is to return all of the data as a string, or upload it to something like an Amazon S3 bucket.

PR #416, opened 2024-09-30
414 Be able to download files via custom forms

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_admin/issues/414

Remaining tasks:

- [x] Allow file downloads via custom forms
- [x] Add tests
- [x] Add docs

For a future PR:

- Add a streaming file response option for large files


piccolo_admin/example.py
```diff
@@ -399,6 +403,14 @@ def validate_email(cls, v):
         return v
 
 
+class DownloadMoviesModel(BaseModel):
+    director_name: str
```
sinisaos, line 407: Should we add a field validator, that director name must not be empty? Something like this.
```python
class DownloadMoviesModel(BaseModel):
    director_name: str

    @field_validator("director_name")
    def validate_director_name(cls, v):
        if v == "":
            raise ValueError("it cannot be empty")
        return v
```
Without that if an empty string is entered in the form, Piccolo admin downloads an empty `csv` file

piccolo_admin/example.py
```diff
@@ -399,6 +403,14 @@ def validate_email(cls, v):
...
+
+
+class DownloadScheduleModel(BaseModel):
+    date: datetime.date
```
sinisaos, line 411: The date field is displayed as a regular string field. I think we should add `v-bind:format="property.format"` to `NewForm` [InputField](https://github.com/piccolo-orm/piccolo_admin/blob/master/admin_ui/src/components/NewForm.vue#L8-L17). After that, the datetime picker field is displayed.

sinisaos:
@dantownsend I think this is a great feature. I left a couple of comments. Maybe it's for some other PR, but now I noticed that Pydantic validation errors not displayed well in UI.

error

dantownsend:
> @dantownsend I think this is a great feature. I left a couple of comments. Maybe it's for some other PR, but now I noticed that Pydantic validation errors not displayed well in UI.
> 
> error

That's interesting - definitely needs improvement! Thanks for highlighting it.

piccolo_admin/example.py
dantownsend: Good point

sinisaos:
@dantownsend Our `parseErrorResponse` utility function works fine, but the problem was in the `responseType blob` in the axios post request. When we use `responseType blob`, we need to use the [text() method to read the content from the Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob#extracting_data_from_a_blob). We need to replace this part of code 

https://github.com/piccolo-orm/piccolo_admin/blob/0132344e752bf55be79530b18982947d9638c233/admin_ui/src/components/FormAdd.vue#L122-L127

with this
```javascript
if (axios.isAxiosError(error) && error.response) {
    const responseData =
        error.request.responseType === "blob" &&
        error.response.data instanceof Blob &&
        error.response.data.type &&
        error.response.data.type
            .toLowerCase()
            .indexOf("json") != -1
            ? JSON.parse(await error.response.data.text())
            : error.response.data
    this.errors = parseErrorResponse(
        responseData,
        error.response.status
    )
}
```
After that we displayed the error messages nicely as before. 

error


Hope that helps.

dantownsend:
> @dantownsend Our `parseErrorResponse` utility function works fine, but the problem was in the `responseType blob` in the axios post request. When we use `responseType blob`, we need to use the [text() method to read the content from the Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob#extracting_data_from_a_blob). We need to replace this part of code
> 
> https://github.com/piccolo-orm/piccolo_admin/blob/0132344e752bf55be79530b18982947d9638c233/admin_ui/src/components/FormAdd.vue#L122-L127
> 
> with this
> 
> ```js
> if (axios.isAxiosError(error) && error.response) {
>     const responseData =
>         error.request.responseType === "blob" &&
>         error.response.data instanceof Blob &&
>         error.response.data.type &&
>         error.response.data.type
>             .toLowerCase()
>             .indexOf("json") != -1
>             ? JSON.parse(await error.response.data.text())
>             : error.response.data
>     this.errors = parseErrorResponse(
>         responseData,
>         error.response.status
>     )
> }
> ```
> 
> After that we displayed the error messages nicely as before.
> 
> error
> 
> Hope that helps.

That's really helpful - thanks!

I wasn't sure what negative effects `responseType blob` would have - it's required for binary files to download correctly.

piccolo_admin/example.py
dantownsend: Done

sinisaos:
> I wasn't sure what negative effects `responseType blob` would have - it's required for binary files to download correctly.

@dantownsend Exactly, `responseType blob` is required to download binaries correctly. With the changes from the previous comment, we downloaded the files correctly and formatted the error messages correctly.

dantownsend:
OK, I've added the blob -> JSON conversion logic now.

dantownsend:
I think the PR is done now.

It ended up being huge because I broke down `example.py` into multiple files because it was getting extremely large.

By breaking it up, it also means I can more easily include chunks of it in the Sphinx docs using `literalinclude`.

Warning - to run the app locally now, you do `python -m piccolo_admin.example.app`.

piccolo_admin/example/forms/csv.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pydantic import BaseModel
+from starlette.requests import Request
+
+from piccolo_admin.endpoints import FileResponse, FormConfig
+from piccolo_admin.example.tables import Movie
+
+
+class DownloadMoviesModel(BaseModel):
+    director_name: str
```
sinisaos, line 12: Just a reminder. Should we add a field validator, that director name must not be empty? Something like this.

```python
class DownloadMoviesModel(BaseModel):
    director_name: str

    @field_validator("director_name")
    def validate_director_name(cls, v):
        if v == "":
            raise ValueError("it cannot be empty")
        return v
```

Without that if an empty string is entered in the form, Piccolo admin downloads an empty csv file

dantownsend: OK, I forgot about that - will change it now.

dantownsend: Done

dantownsend:
@sinisaos Do you see anything else, or am I good to merge?

I know it's a massive PR, with a lot of noise in it.

dantownsend:
@sinisaos Cool, thanks a lot - I really appreciate it 👍

PR #427, opened 2024-10-18
Dockerised piccolo_admin

dl-lim:
Basic copy of admin_ui in docker.

To add a standalone application in docker next, with .env to configure db connections.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+# Use an official Node.js runtime as a parent image
+FROM node:20
+
+# Set the working directory to /admin_ui
+WORKDIR /admin_ui
+
+# Copy the current directory contents into the container at /admin_ui
+COPY admin_ui .
+
+# Install any needed packages specified in package.json
+RUN npm install --no-audit --no-fund
+
+# Expose port 3000
+EXPOSE 3000
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY admin_ui .
+
+# Install any needed packages specified in package.json
+RUN npm install --no-audit --no-fund
+
+# Expose port 3000
+EXPOSE 3000
+
+# Run npm run dev when the container launches
+CMD ["npm", "run", "dev"]
```
dantownsend, line 17: This solution is OK for development, but for production we don't need to install the npm dependencies etc. When we release `piccolo_admin` on PyPI, we bundle all of the front end assets.

So we just need a Python file which runs `create_admin`, but reads the database values from environment variables.

sinisaos:
@dl-lim I think you should close this PR as we now have a [dedicated repo](https://github.com/piccolo-orm/piccolo-admin-docker) for dockerized Piccolo Admin. Feel free to make changes to that repo. Thanks.

sinisaos:
Closed this PR because we now have a [dedicated repo](https://github.com/piccolo-orm/piccolo-admin-docker) for the dockerized Piccolo admin.

PR #428, opened 2024-11-14
fix wrong method in docs

sinisaos:
Fixing a small mistake around encryption key generation in docs. `XChaCha20Provider.generate_key()` raise `AttributeError: type object 'XChaCha20Provider' has no attribute 'generate_key'` (because `generate_key` is [Fernet method](https://github.com/piccolo-orm/piccolo_api/blob/bc7eefa9d63b018cf45853798acbfd0be0978754/piccolo_api/encryption/providers.py#L112C16-L112C37)). I think it should be https://github.com/piccolo-orm/piccolo_api/blob/bc7eefa9d63b018cf45853798acbfd0be0978754/piccolo_api/encryption/providers.py#L179

### DIFF
--- a/docs/source/mfa/index.rst
+++ b/docs/source/mfa/index.rst
@@ -49,8 +49,8 @@ To generate the encryption key in the above example:
 .. code-block:: pycon
 
     >>> from piccolo_api.encryption.providers import XChaCha20Provider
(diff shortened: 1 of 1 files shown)


dantownsend:
@sinisaos Good catch, thanks!

Issue #430, opened 2025-01-14
UI inconsistencies within forms


Skelmis:
Piggy backing on https://github.com/piccolo-orm/piccolo_admin/issues/134 but further diving into it, the forms UI fails to utilise the data provided via the `schema` route under some scenarios.

### Case one

Code:
```py
class Form(BaseModel):
    end_date: Optional[date] = None
```

Schema:
```json
{
	"end_date": {
		"anyOf": [
			{
				"format": "date",
				"type": "string"
			},
			{
				"type": "null"
			}
		],
		"default": null,
		"title": "End Date"
	}
}
```

Expectation:
UI displays the date picker, but with no value selected by default

Actual:
Empty `text` field

### Case two

Code:
```py
class Form(BaseModel):
    enum: MyEnum
```

Schema:
Includes a $defs to the schema area with the correct data

Expectation:
Drop down with the values

Actual:
1. It fails to put a name above to the text field
2. Empty text field

dantownsend:
Thanks for reporting this - you're right, I can replicate this issue.

Enums are quite tricky, because the OpenAPI schema uses refs, which are kind of a pain. I would like to support these though.

Things like optional date fields not working is definitely a bug.

dantownsend:
I've found the cause of the optional date field bug - will put up a PR.

PR #432, opened 2025-01-14
430 Fix bug with optional fields in custom forms

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_admin/issues/430

### DIFF
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
     playwright:
         name: "Playwright Tests"
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-22.04
         steps:
             - name: "Checkout repository"
               uses: "actions/checkout@v4"
--- a/admin_ui/src/utils.ts
+++ b/admin_ui/src/utils.ts
@@ -1,4 +1,9 @@
-import { type Schema, type OrderByConfig, getType } from "@/interfaces"
+import {
+    type Schema,
+    type OrderByConfig,
+    getType,
+    isNullable
+} from "@/interfaces"
 import router from "./router"
 import moment from "moment"
 
@@ -140,6 +145,10 @@ export function convertFormValue(params: {
     if (value == "null") {
         value = null
     } else if (property.extra?.nullable && value == "") {
+        // TODO We can potentially remove this in the future - isNullable does
+        // what we need.
+        value = null
+    } else if (isNullable(property) && value == "") {
         value = null
     } else if (getType(property) == "array") {
         value = JSON.parse(String(value))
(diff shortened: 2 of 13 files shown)


.github/workflows/playwright.yaml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
     playwright:
         name: "Playwright Tests"
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-22.04
```
dantownsend, line 12: It looks like Playwright isn't running out of the box on Ubuntu 24.04 at the moment due to some missing dependencies.

https://github.com/microsoft/playwright/issues/30368

piccolo_admin/endpoints.py
```diff
@@ -680,7 +680,7 @@ def __init__(
 
         private_app.add_route(
             path="/change-password/",
-            route=change_password(
+            route=change_password(  # type: ignore
```
dantownsend, line 683: I'm pretty sure the Starlette type annotations are wrong here - we're just passing in a `HTTPEndpoint`.

piccolo_admin/endpoints.py
```diff
@@ -786,7 +786,7 @@ def __init__(
 
         public_app.add_route(
             path="/logout/",
-            route=session_logout(session_table=session_table),
+            route=session_logout(session_table=session_table),  # type: ignore
```
dantownsend, line 789: See above about type annotations.

admin_ui/src/utils.ts
```diff
@@ -140,6 +145,10 @@ export function convertFormValue(params: {
     if (value == "null") {
         value = null
     } else if (property.extra?.nullable && value == "") {
+        // TODO We can potentially remove this in the future - isNullable does
+        // what we need.
+        value = null
+    } else if (isNullable(property) && value == "") {
```
dantownsend, line 151: This is problematic with the current version of Piccolo - in the Pydantic model for tables, they default to optional, unless `required=True`.

This PR fixes it:

https://github.com/piccolo-orm/piccolo/pull/1141

But need to decide whether it's a good idea or not.

Skelmis (approved):
Tested, lgtm

admin_ui/src/utils.ts
Skelmis: I don't really understand what the implication here is, maybe @sinisaos has some thoughts?

sinisaos:
I also tested this PR against the latest Piccolo ORM PR and in my opinion everything works fine.

admin_ui/src/utils.ts
dantownsend: Thanks both for testing it. 

I'll change this slightly, so it's completely backwards compatible - i.e. the current behaviour for tables, and for custom forms it uses `isNullable`.

dantownsend:
I made this backwards compatible, so it doesn't require any changes to `create_pydantic_model` - it gives us some more time to think it through. Thanks both!

PR #433, opened 2025-01-16
Update contributing documentation for local backends

Skelmis:
The backend requires the project to be built for `dist` files to be in place. This PR just makes that a touch clearer


dantownsend:
Thanks for this. The PR is failing due to some unrelated problems with the CI. I'll fix it, then merge in this branch.

PR #434, opened 2025-01-16
Support Enum types within Forms

Skelmis:
My attempt at supporting Enum's within Forms.

Note this is my first time working with Vue and as such I'm trying to figure things out and just get them working. I'd love feedback and improvements.

#430


admin_ui/src/components/NewForm.vue
```diff
@@ -23,7 +37,16 @@
 <script lang="ts">
 import { defineComponent, type PropType } from "vue"
 import InputField from "./InputField.vue"
-import { type Schema, getType } from "@/interfaces"
+import { Enum, Choices, type Schema, getType } from "@/interfaces"
+
+export function enumToChoices(data: Enum): Choices {
```
Skelmis, line 42: Trying to figure out how to expose this to the template at the moment...

Skelmis: Got it.

Skelmis:
So PoC form:
```py
from enum import Enum

from pydantic import BaseModel
from starlette.requests import Request

from piccolo_admin.endpoints import FormConfig


class Test(Enum):
    ONE = "ONE"
    TWO = "TWO"
    THREE = "THREE"

class CalculatorModel(BaseModel):
    number_1: int
    number_2: int
    enum: Test


def calculator(request: Request, data: CalculatorModel):
    """
    A very simple example of a form which adds numbers together.
    """
    return f"The answer is {data.number_1 + data.number_2} with enum {data.enum}."


FORM = FormConfig(
    name="Calculator",
    pydantic_model=CalculatorModel,
    endpoint=calculator,
    description=("Adds two numbers together."),
)
```


Produces:
image

sinisaos:
@Skelmis There was a [PR to enable Python Enums in custom forms](https://github.com/piccolo-orm/piccolo_admin/pull/391). I used Piccolo `choices` because Piccolo Admin and Piccolo API doesn't understand references through `"$ref"`. Everything worked (video [here](https://github.com/piccolo-orm/piccolo_admin/issues/386#issuecomment-2080395051) or video [here](https://github.com/piccolo-orm/piccolo_admin/issues/386#issuecomment-2079787739)) and was easy for the user to use, but I'm closing that PR because it was decided that we need to move to a native OpenAPI approach (that would be best). Piccolo Admin and Piccolo API do not currently use complex references and to work natively we need to make changes to both repos

Skelmis:
What changes are required within the API? Possibly I've overlooked something but I don't feel like it may

sinisaos:
@Skelmis `create_pydantic_model` does not use `definitions` and `$ref `(as in OpenAPI for `enums`). [Here](https://github.com/piccolo-orm/piccolo/pull/469) and [here](https://github.com/piccolo-orm/piccolo/issues/331) are possible ideas of changes that should be made, but you have to wait @dantownsend to see how it should be done.

Skelmis:
Interesting, I wonder then how $refs are currently propgrating then given this draft is essentially working. I might take a poke later

sinisaos:
Ah my mistake, sorry. I forgot that custom forms using Pydantic `BaseModel` do not `create_pydantic_model` (like table forms) and the `$ref` exists in the json schema. After that you parse the properties (I tried pretty much the same thing in a closed PR, but I used Piccolo `choices`). Sorry again.

dantownsend:
@Skelmis Thanks for taking a look at this, looks like some great progress!

I don't mind us changing `create_pydantic_model` so it returns choices as `$ref`. At the moment we add it as an extra attribute to the Pydantic JSON schema:

https://github.com/piccolo-orm/piccolo/blob/965c5b3ef2311a0ecb39a09341d5a6ebabc25b5d/piccolo/utils/pydantic.py#L280C37-L280C53

Skelmis:
I actually prefer how it currently works because then you get a display value and the actual value versus $ref which at face value is just the raw value

Skelmis:
I'm currently getting the following error, could either of you give me a hand resolving it?
```
src/components/NewForm.vue:13:37 - error TS2722: Cannot invoke an object which is possibly 'undefined'.

13                     v-bind:choices="etc(schema['$defs'][Number(property['$ref']?.split(/[//]+/).pop())])"
                                       ~~~


Found 1 error in src/components/NewForm.vue:
```

Skelmis:
Closing in favor of #436

PR #436, opened 2025-01-25
add enum support to custom forms

sinisaos:
An alternative approach to #434 . It uses Piccolo choices without complex parsing definitions in the Vue frontend.


sinisaos:
@Skelmis I can't add you as a reviewer, but I'm interested in your thoughts on this approach of using Python enums in custom forms. Thanks in advance.

Skelmis:
It looks like a much simpler implementation which I love. I'll try test it out within a week

sinisaos:
@Skelmis Great. Thanks

Skelmis (commented):
Tested and it works as is, my only issue is that using the example we are accessing an untyped field.

I.e. 
`data: NewStaffModel` yet we need `data.permissions` which doesnt exist

Users could instead do something like this:
```py
class NewStaffFormModel(BaseModel):
    username: str
    email: EmailStr
    superuser: bool

class NewStaffModel(NewStaffFormModel):
    permissions: Permission

def new_staff_endpoint(request: Request, data: NewStaffModel) -> str:
    print(type(data.permissions))
    return "A new staff member has been successfully created."


FORM = FormConfig(
    name="Enum form",
    pydantic_model=NewStaffFormModel,
    endpoint=new_staff_endpoint,
    description="Make a enum form.",
    choices={"permissions": Permission},
    form_group="Text forms",
)
```
But even in that case `type(data.permissions)` is `str` instead of `Permissions`.

Do you think it'd be possible to try return Enum items as the relevant enum in this case? Or should users do the enum conversion themselves

sinisaos:
@Skelmis Thank you for your time to try this out.

> Do you think it'd be possible to try return Enum items as the relevant enum in this case? Or should users do the enum conversion themselves

Yes, you are right. I think the easiest way, as you say, is for users to do the enum conversion themselves. Something like this

```python
def new_staff_endpoint(request: Request, data: NewStaffModel) -> str:
    # data.permissions = Permission(int(data.permissions)) # for int enum
    data.permissions = Permission(data.permissions) # for str enum
    print(type(data.permissions)) # prints correct type <enum 'Permission'>
    return "A new staff member has been successfully created."
```

Skelmis:
Seems reasonable for me. Thoughts @dantownsend ?

Skelmis (approved):

Skelmis:
Hrmm perhaps just realised we should likely add docs haha. Will see if I find time sometime to try write some up

sinisaos:
@Skelmis Yes, you are right, but I don't see the point of writing docs because PR is not reviewed or accepted. We can add a usage example to the docs later, if PR is accepted.

Skelmis:
Fair enough

Skelmis:
@dantownsend what's your view on this pr? Docs before pr, or pr then docs on approved content? 

Either way keen to get this moving, would love this feature

Skelmis:
> @dantownsend what's your view on this pr? Docs before pr, or pr then docs on approved content? 
> 
> Either way keen to get this moving, would love this feature

@dantownsend waiting on your response to move this PR forward

dantownsend:
Sorry for the delay on this. I was torn about which approach to take:

1. Should we make our front end understand the refs in the OpenAPI schema.
1. Should we add the `extra` attributes (like this in this PR).

There was also some discussion about not using OpenAPI schemas any more to drive the front end (because Litestar doesn't allow the `extra` attributes).

This PR might be the best approach for now though.

piccolo_admin/endpoints.py
```diff
@@ -395,6 +396,7 @@ def __init__(
             t.Union[FormResponse, t.Coroutine[None, None, FormResponse]],
         ],
         description: t.Optional[str] = None,
+        choices: t.Optional[t.Dict[str, t.Any]] = None,
```
dantownsend, line 399: Rather than passing the choices in, can't we just look at the enum value on the pydantic model itself (i.e. if the default value for the field is an enum then use it to generate the choices).

sinisaos: Do you think something like this?

<details>
<summary>Code changes</summary>

```python
def __init__(
      self,
      name: str,
      pydantic_model: t.Type[PydanticModel],
      endpoint: t.Callable[
          [Request, PydanticModel],
          t.Union[FormResponse, t.Coroutine[None, None, FormResponse]],
      ],
      description: t.Optional[str] = None,
      form_group: t.Optional[str] = None,
  ):
      self.name = name
      self.pydantic_model = pydantic_model
      self.endpoint = endpoint
      self.description = description
      self.form_group = form_group
      self.slug = self.name.replace(" ", "-").lower()
      for (
          field_name,
          field_value,
      ) in self.pydantic_model.model_fields.items():
          if isinstance(field_value.annotation, enum.EnumType):
              # update model_fields, field annotation and
              # rebuild the model for the changes to take effect
              pydantic_model.model_fields[field_name] = Field(
                  json_schema_extra={
                      "extra": {
                          "choices": convert_enum_to_choices(
                              field_value.annotation
                          )
                      }
                  },
              )
              pydantic_model.model_fields[field_name].annotation = str
              pydantic_model.model_rebuild(force=True)
```

</details>

Usage would be

<details>
<summary>Usage example</summary>

```python
# An example of using Python enum in custom forms
class Permission(str, enum.Enum):
    admissions = "admissions"
    gallery = "gallery"
    notices = "notices"
    uploads = "uploads"


class NewStaffModel(BaseModel):
    username: str
    email: EmailStr
    superuser: bool
    permissions: Permission


def new_staff_endpoint(request: Request, data: NewStaffModel) -> str:
    # We need to do the enum type conversion ourselves like this:
    # data.permissions = Permission(int(data.permissions)) # for int enum
    # data.permissions = Permission(data.permissions)  # for str enum
    print(data)
    return "A new staff member has been successfully created."


FORM = FormConfig(
    name="Enum form",
    pydantic_model=NewStaffModel,
    endpoint=new_staff_endpoint,
    description="Make a enum form.",
    form_group="Text forms",
)
```
</details>

piccolo_admin/endpoints.py
```diff
@@ -403,6 +412,24 @@ def __init__(
...
+                pydantic_model.model_fields[field_name] = Field(
+                    json_schema_extra={
+                        "extra": {
+                            "choices": convert_enum_to_choices(
+                                field_value.annotation
+                            )
+                        }
+                    },
+                )
+                pydantic_model.model_fields[field_name].annotation = str
```
dantownsend, line 431: Can't we just leave the annotation as what it was originally?

piccolo_admin/translations/data.py
```diff
@@ -1098,7 +1098,7 @@
 FARSI = Translation(
     language_name="FARSI",
     language_code="fa",
-    translations = {
+    translations={
```
dantownsend, line 1101: Thanks for fixing these - I'm going to pull it out into a separate PR.

piccolo_admin/endpoints.py
```diff
@@ -403,6 +412,24 @@ def __init__(
         self.description = description
         self.form_group = form_group
         self.slug = self.name.replace(" ", "-").lower()
+        for (
+            field_name,
+            field_value,
+        ) in self.pydantic_model.model_fields.items():
+            if isinstance(field_value.annotation, ENUMTYPE):
```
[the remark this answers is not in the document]
sinisaos: Note that the field value annotation is of type `EnumType` (before Python 3.11 it was `EnumMeta`), not of **type enum.Enum** and we need to compare the object with `ENUMTYPE` not `enum.Enum`. It's practically the same to use `issubclass(type(field_value.annotation), ENUMTYPE)` or if `isinstance(field_value.annotation, ENUMTYPE)`, but I find the `isinstance` option a bit cleaner.

piccolo_admin/endpoints.py
sinisaos: We need to reassign the annotation type because if we don't, that annotation is `None` and that causes a form validation error. I'm using the string type because if we reassign annotation to the `Enum`, frontend don't work (working with strings because choices `display_value` and `value` are strings, and choices does not work with `Enum`). In the whole PR I tried to use Piccolo choices so that there would be as few changes as possible on the frontend. If you don't like the whole concept, feel free to delete this PR.

dantownsend: Ah, yes - I forgot that the front end would get confused by the open api schema is we leave the enum in there.

piccolo_admin/endpoints.py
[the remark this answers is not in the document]
sinisaos: Yes, it works. Great.

docs/source/custom_forms/index.rst
```diff
@@ -39,6 +39,16 @@ Here's a more advanced example where we send an email, then return a string:
...
 
+``Enum``
+--------
+
+Custom forms support ``Enum`` type. Here's a example:
+
+.. literalinclude:: ../../../piccolo_admin/example/forms/enum.py
+
+.. warning::
+    We need to do the ``Enum`` type conversion from the form data ourselves.
```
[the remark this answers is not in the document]
sinisaos: @Skelmis Thank you, I am not a native English speaker and help with documents is always welcome in my case. Thanks again.

Skelmis (approved):

Skelmis:
@dantownsend do you have any updates regarding this PR? Is it blocked as a result of technical basis or more so just personal time constraints

sinisaos:
I'm closing this PR just to clean up my open requests in the repo. Since it has been open for quite a while, I'll remove it from the list to keep things tidy. Feel free to use it if needed. Thanks.

Issue #440, opened 2025-03-19
Frontend doesn't take into account the `page_size` parameter


metakot:
```python
from piccolo_admin.endpoints import create_admin

admin_app = create_admin(
    tables=[SomeTable],
    page_size=50,
)
```

![Image](https://github.com/user-attachments/assets/9427af46-fd8e-4530-b60c-19469b78ad8f)

The function [create_admin](https://github.com/piccolo-orm/piccolo_admin/blob/309932fb91ccef7ede8c7fa164da8a86e060f198/piccolo_admin/endpoints.py#L1190) takes a `page_size` parameter, which later provided to [piccolo CRUD](https://github.com/piccolo-orm/piccolo_admin/blob/309932fb91ccef7ede8c7fa164da8a86e060f198/piccolo_admin/endpoints.py#L593), but it's not getting to the frontend vue part and the admin page still uses the default `15` pages which is hardcoded elsewhere, I think.

dantownsend:
You're right - good catch. The value is available in the API:

```
GET http://127.0.0.1:8000/api/tables/movie/count/

{"count":17,"page_size":30}
```

But the front end isn't using it. This is the component:

https://github.com/piccolo-orm/piccolo_admin/blob/master/admin_ui/src/components/ChangePageSize.vue

PR #441, opened 2025-03-19
440 Frontend doesn't take into account the `page_size` parameter

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_admin/issues/440

### DIFF
--- a/admin_ui/src/store.ts
+++ b/admin_ui/src/store.ts
@@ -23,7 +23,7 @@ export default createStore({
         currentTableName: undefined,
         darkMode: false,
         filterParams: {} as { [key: string]: any },
-        pageSize: 15,
+        pageSize: 0,
         rowCount: 0,
         rows: [],
         schema: undefined as i.Schema | undefined,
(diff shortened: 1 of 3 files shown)


sinisaos:
@dantownsend I don't know if this helps, but the Playwright test passes if we change the response url from this

 https://github.com/piccolo-orm/piccolo_admin/blob/309932fb91ccef7ede8c7fa164da8a86e060f198/e2e/test_sort_columns.py#L65-L70

to this

```python
with page.expect_response(
    lambda response: "/tables/sorted_columns/?" in response.url
):
```

dantownsend:
> @dantownsend I don't know if this helps, but the Playwright test passes if we change the response url from this
> 
> https://github.com/piccolo-orm/piccolo_admin/blob/309932fb91ccef7ede8c7fa164da8a86e060f198/e2e/test_sort_columns.py#L65-L70
> 
> to this
> 
> ```python
> with page.expect_response(
>     lambda response: "/tables/sorted_columns/?" in response.url
> ):
> ```

@sinisaos Thanks - I'll give that a go!

sinisaos:
@dantownsend @hoosnick The code from this PR actually doesn't work and that's why the Playwright test fails. After #468 and #469 are merged, I'll create a PR to add 
1. `__page` and `__page_size` to the query parameters (for proper filtering with query parameters from the URL)
2. I'll also add `__order` to the query parameters because without it the ordering doesn't work correctly (for proper filtering with query parameters from the URL)
3. And finally I'll add code for the `page_size` parameter (instead of this PR) which will be done based on user `page_size` param (if parameter does not exists in config, default stays 15). 

I already have code for all that.

PR #446, opened 2025-05-20
added farsi translation

parsa-asgari:
# Notes
- Adds Farsi language translation


dantownsend:
@parsa-asgari Looks great - thanks!

PR #449, opened 2025-06-06
drop support for Python 3.8 and add Python 3.13

sinisaos:
Drop support for Python 3.8 and add support for Python 3.13


dantownsend:
@sinisaos Looks great - thanks!

PR #451, opened 2025-07-23
Add italian translation

Ott8v:
Adds Italian language translation


piccolo_admin/translations/data.py
```diff
@@ -1190,6 +1273,7 @@
     UKRAINIAN,
     WELSH,
     FARSI,
+    ITALIAN
```
sinisaos, line 1276:
```suggestion
    ITALIAN,
```
Missing comma. Can you please also run `./scripts/lint.sh` (from the root directory) to run the linters (to fix whitespace if there are any issues).Thanks

Issue #463, opened 2025-12-16
Performance degredation on large datasets


Skelmis:
When using the Admin UI for large datasets (Millions of rows), there is a noticeable hang time before the UI either crashes or loads. Reloading the page when loaded on a table with a small dataset takes a few seconds where as a dataset of a million takes half a minute on my machine. I imagine larger datasets would be worse.

It was initially reported in the [discord here](https://discord.com/channels/774888669332766750/774888669332766753/1450337010521735199) but I can also reproduce it.

---

Initial profiling of a pip installed version of the UI point to this:

<img width="1844" height="570" alt="Image" src="https://github.com/user-attachments/assets/d05c9230-78c7-4cf1-a602-142d10d04300" />

I dont know why its chonking the GC so much if its paginated, which it is by default but figured it could be useful context to add

sinisaos:
@Skelmis The main issue with large datasets is [LimitOffset](https://piccolo-api.readthedocs.io/en/latest/crud/piccolo_crud.html#pagination) pagination (used in `PiccoloCRUD`). For `LimitOffset` pagination, the database has to scan and discard all rows up to `OFFSET`, which leads to poor performance on large datasets where the offset is large. Cursor pagination has much better performance ([as discussed here](https://github.com/piccolo-orm/piccolo_api/issues/15) and there is a [separate package](https://github.com/piccolo-orm/piccolo_cursor_pagination) that does not work with Piccolo API and Piccolo Admin), but there are [limitations](https://github.com/piccolo-orm/piccolo_cursor_pagination?tab=readme-ov-file#limitations). Not being able to go to a specific page is, I think, a big problem and limitation for any admin dashboard. I don't think there's much we can do to solve performance issues with large datasets.

dantownsend:
Yeah, that's strange.

I tried it locally with 1 million rows, using this (you need the latest from master branch for this to work):

```
python -m piccolo_admin.example.app --inflate=1000000 --engine=postgres
```

Apart from it taking about 5 minutes to start up, because it was generating so much fake data, it ran surprisingly OK. My work machine is pretty powerful though, and has 36 GB of RAM.

The tables I was using were also fairly simple. I wonder if there's something about the tables which can make it worse.

Weirdly, one of the things which was really lagging at first was the pagination selector. It's just a HTML select widget, but it had thousands of values in it. For a million rows, and the default 15 rows per page, that's over 60k elements!

[...]

PR #468, opened 2025-12-19
Better pagination for large datasets

sinisaos:
Related to #463


dantownsend (commented):
Looks good 👍 

Can you use `text-align: center` on the input box?

Also, we need to detect if users click the up and down buttons on the input box - at the moment, it doesn't seem to move to the next/previous page.

Screenshot 2025-12-19 at 17 52 18

Issue #466, opened 2025-12-18
Reusable search filter using URLs


abhishek-compro:
Currently when I filter for anything, the URL stays the same (example: http://localhost:8000/admin/#/users/). This is a big drawback as I will have to specify the filters separately along with the URL that I am sharing with anyone.

What I would want is something like: http://localhost:8000/admin/#/users/?name__contains=john&id__gt=1321

sinisaos:
This could be achieved by adding something like this
```js
// adding query params to url
this.$router.replace({
    path: this.$router.path,
    query: {
        ...json
    }
})
```
to here in `RowFilter.vue`

https://github.com/piccolo-orm/piccolo_admin/blob/7c69df9ca361479b0c150f7359f6d332dc2534cc/admin_ui/src/components/RowFilter.vue#L89-L91

To clear query parameters in `clearFilters` function we can add something like this

```js
// cleaning query params from url
this.$router.replace({
    path: this.$router.path,
    query: {}
})
```
Result is

<img width="1432" height="366" alt="Image" src="https://github.com/user-attachments/assets/1946171a-aeaf-4aad-97f0-30dd6db5e4ce" />

sinisaos:
Close this issue which is actually a duplicate of #42 . Both issues should be resolved with #469

PR #469, opened 2025-12-19
Filter params in url

sinisaos:
Related to #466


dantownsend (commented):
Thanks for this. It's cool how it syncs the filter values to the URL params.

But I think the missing piece is the reverse - so if you copy the URL, and post it into a new tab, it should then take the URL params, and populate the filter sidebar.

sinisaos:
> @sinisaos This looks great, thanks! I've done some testing, and it seems to work for all field types now?

Yes. It should work for all field types.

> The only thing I'd like to add in a future PR is also syncing the page number to the query params too.

Yes, that would be nice. I can try to do that, but you need to merge this PR and the pagination PR first, because there's no point in trying to make those changes now, if the pagination is going to change.

sinisaos:
I'm closing this PR just to clean up my open requests in the repo. Since it has been open for quite a while, I'll remove it from the list to keep things tidy. Feel free to use it if needed. Thanks.

